PR #2, opened 2023-07-15
Add local video track support

gongxinheng:
Add local video track support and publishing hue video track example

### DIFF
--- a/include/livekit/ffi_client.h
+++ b/include/livekit/ffi_client.h
@@ -63,16 +63,16 @@ class FfiClient {
 
 struct FfiHandle {
  public:
-  FfiHandle(FfiHandleId handle) : handle_(handle) {}
+  FfiHandle(FfiHandleId handleId) : handleId_(handleId) {}
 
   FfiHandle(FfiHandle&& handle) {
-    handle_ = handle.handle_;
-    handle.handle_ = INVALID_HANDLE;
+    handleId_ = handle.handleId_;
+    handle.handleId_ = INVALID_HANDLE;
   }
 
   ~FfiHandle() {
-    if (handle_ != INVALID_HANDLE) {
-      assert(livekit_ffi_drop_handle(handle_));
+    if (handleId_ != INVALID_HANDLE) {
+      assert(livekit_ffi_drop_handle(handleId_));
     }
   }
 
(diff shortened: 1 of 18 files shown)


src/participant.cpp
```diff
@@ -0,0 +1,74 @@
...
+  if (!room) {
+    throw std::runtime_error("room is not available");
+  }
+
+  proto::FfiRequest request{};
+  proto::PublishTrackRequest* publishTrackRequest =
+      request.mutable_publish_track();
+  publishTrackRequest->set_track_handle(
+      track->ffiHandle_.GetHandleId());
+  *publishTrackRequest->mutable_options() = options;
```
theomonnom, line 49: Seems like your forgot to put the participant_handle inside the request

include/livekit/participant.h
```diff
@@ -0,0 +1,65 @@
...
+  LocalParticipant(proto::ParticipantInfo&& info, std::shared_ptr<Room> room)
+      : Participant(std::move(info)), room_(room) {}
+
+  ~LocalParticipant();
+
+  void PublishTrack(std::shared_ptr<Track> track,
+                    const proto::TrackPublishOptions& options);
+
+ private:
+  std::condition_variable cv_;  // Should we block?
```
theomonnom, line 55: I don't think we should have a blocking API.
Maybe it is OK to add callback pointers inside the arguments?

include/livekit/participant.h
```diff
@@ -0,0 +1,65 @@
...
+  uint64_t publishAsyncId_;
+  FfiClient::ListenerId listenerId_{0};
+  std::unique_ptr<proto::PublishTrackCallback> publishCallback_;
+  std::weak_ptr<Room> room_;
```
theomonnom, line 59: With the new FFI version, it is no longer required to keep a reference to the room here (request don't need it)

include/livekit/ffi_client.h
```diff
@@ -17,57 +17,79 @@
...
+  ~FfiHandle() {
+    if (handleId_ != INVALID_HANDLE) {
+      assert(livekit_ffi_drop_handle(handleId_));
+    }
+  }
 
-        FfiClient();
-        ~FfiClient() = default;
+  FfiHandle(FfiHandle const&) = delete;
+  FfiHandle& operator=(FfiHandle const&) = delete;
```
theomonnom, line 80: Maybe we should add a unit tests that use the operator= (You said you saw issues related to dropping twice an handle)

theomonnom:
Open to feedbacks wrt exposing events to the users (Should use do a virtual class? Event polling)
Cc @cloudwebrtc

PR #5, opened 2025-11-06
Revive the client-sdk-cpp repo and make it compilable and runnable

xianshijing-lk:
The PR does:
- Fixed the CMakelist.txt to make it compilable
- Fixed README.md
- Fixed the example
- Added build.sh to make it easier to clean | build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,31 +104,6 @@ add_library(livekit
   src/ffi_client.cpp
   src/room.cpp
 )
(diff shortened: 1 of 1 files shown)


ladvoc:
@xianshijing-lk, it looks like the CI workflow doesn't work anymore—should we address that here or in a separate PR? Also, I see this repo builds the Rust SDK from source rather than pulling in a prebuilt dylib. Do you have any thoughts on which you prefer?

PR #14, opened 2025-12-10
Sxian/clt 2342/implement the remaining room events

xianshijing-lk:
This PR implements the remaining room events
and separate the room_event_types.h and room_delegate.h

It also adds the comments / docs to the room_delegates.h and room_event_types.h

### DIFF
--- a/include/livekit/room_event_types.h
+++ b/include/livekit/room_event_types.h
@@ -106,29 +106,6 @@ enum class DisconnectReason {
 };
 
 /**
- * A chat message associated with the room.
- */
-struct ChatMessageData {
-  /** Unique ID of the message. */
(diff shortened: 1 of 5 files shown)


include/livekit/room_event_types.h
```diff
@@ -0,0 +1,808 @@
...
+  UserRejected,
+  SipTrunkFailure,
+  ConnectionTimeout,
+  MediaFailure
+};
+
+/**
+ * A chat message associated with the room.
+ */
+struct ChatMessageData {
```
ladvoc: note: Chat messages are a deprecated in favor of data streams. Not sure if we want to expose in new code.

PR #16, opened 2025-12-13
Data Stream features and example

xianshijing-lk:
This PR:
Implement the data stream and hook it up to the room
remove the publishTranscription and its relevant code
Implement the data stream example, which will send chat msgs and file to the other participants.
update the readme with the new example

### DIFF
--- a/examples/simple_data_stream/main.cpp
+++ b/examples/simple_data_stream/main.cpp
@@ -187,7 +187,7 @@ void handleWelcomeImage(std::shared_ptr<livekit::ByteStreamReader> reader,
 } // namespace
 
 int main(int argc, char *argv[]) {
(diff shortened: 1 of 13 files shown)


ladvoc:
One important note: this implementation uses the deprecated FFI methods:
- `send_stream_header`
- `send_stream_chunk`
- `send_stream_trailer`

We have since added [high-level methods](https://github.com/livekit/rust-sdks/blob/6464e2e9bdb92130bf77f1ca2750586334cd6184/livekit-ffi/protocol/ffi.proto#L132) which handle all the protocol-level details on the Rust side and handle error/edge cases that the Python implementation doesn't. Currently, these new methods have only been adopted in Unity. I don't think we need to upgrade this right away since it works now—just wanted to make a note of it.

PR #20, opened 2025-12-23
generate docs from the publit folders

xianshijing-lk:


ladvoc (approved):
LGTM ✅. If you want it to auto deploy to GitHub pages, there's a workflow for that in the ESP32 repo.

PR #35, opened 2026-01-09
libprotobuf and abseil_dll libs are missing on windows, fix them

xianshijing-lk:
The problem is that install(IMPORTED_RUNTIME_ARTIFACTS does not install the relevant .dll.lib into the lib/ folder.

This PR is trying to fix it


xianshijing-lk:
Those install() cmd will be triggered when we are making the release, and I don't have a windows machine to test. 

I have to land it as is, and make an test release to see if that fixes the issue.

PR #59, opened 2026-02-15
Ingest and publish yuv frames

haffi96:
A simple example to injest yuv frames and publish to livekit

### DIFF
--- a/examples/publish_yuv_source/yuv_source.cpp
+++ b/examples/publish_yuv_source/yuv_source.cpp
@@ -43,14 +43,16 @@ namespace {
 socket_t connectTcp(const std::string &host, std::uint16_t port) {
 #ifdef _WIN32
   WSADATA wsa;
-  if (WSAStartup(MAKEWORD(2, 2), &wsa) != 0) return INVALID_SOCKET_VALUE;
+  if (WSAStartup(MAKEWORD(2, 2), &wsa) != 0)
+    return INVALID_SOCKET_VALUE;
 #endif
   struct addrinfo hints = {}, *res = nullptr;
   hints.ai_family = AF_UNSPEC;
   hints.ai_socktype = SOCK_STREAM;
   std::string portStr = std::to_string(port);
   if (getaddrinfo(host.c_str(), portStr.c_str(), &hints, &res) != 0) {
-    std::cerr << "YuvSource: getaddrinfo failed for " << host << ":" << port << "\n";
+    std::cerr << "YuvSource: getaddrinfo failed for " << host << ":" << port
+              << "\n";
     return INVALID_SOCKET_VALUE;
   }
   socket_t fd = socket(res->ai_family, res->ai_socktype, res->ai_protocol);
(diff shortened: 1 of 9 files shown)


examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+#endif
+
+namespace publish_yuv {
+
+namespace {
+
+socket_t connectTcp(const std::string &host, std::uint16_t port) {
+#ifdef _WIN32
+  WSADATA wsa;
+  if (WSAStartup(MAKEWORD(2, 2), &wsa) != 0)
```
xianshijing-lk: I am not that familiar with the windows API. Do we need to clean up / release the resource explicitly in Stop() as we are calling WSAStartup() in connectTcp() function ?

examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  std::cout << "YuvSource: connected to " << host_ << ":" << port_ << " ("
+            << width_ << "x" << height_ << "@" << fps_
+            << "fps, frame=" << frame_size_ << " bytes)\n";
+
+  auto t0 = std::chrono::steady_clock::now();
+
+  while (running_.load()) {
+    std::vector<std::uint8_t> frame(frame_size_);
```
xianshijing-lk: We'd better not allocate a new vector of frame() for every run inside the loop.

Move it out, and reserve the frame_size_, like 
std::vector<std::uint8_t> buf(frame_size_);

If the frame_size_ can change / increased over time, ideally we can allocate a big enough buffer, otherwise just double the vector if frame_size_ is bigger the size of frames

examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+    std::size_t filled = 0;
+    while (filled < frame_size_ && running_.load()) {
+#ifdef _WIN32
+      int n = recv(fd, reinterpret_cast<char *>(frame.data() + filled),
```
xianshijing-lk: will recv() block the application forever in some cases? 

I wonder if we can add a recv timeout or allow the application to interrupt with ctrl + c ?

examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+                   static_cast<int>(frame_size_ - filled), 0);
+#else
+      ssize_t n = recv(fd, frame.data() + filled, frame_size_ - filled, 0);
+#endif
+      if (n <= 0) {
```
xianshijing-lk: Log it ? I wonder what use case recv will return negative values

examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+      break;
+
+    std::int64_t ts_us = std::chrono::duration_cast<std::chrono::microseconds>(
+                             std::chrono::steady_clock::now() - t0)
+                             .count();
+    if (callback_) {
+      YuvFrame out;
+      out.data = std::move(frame);
+      out.timestamp_us = ts_us;
+      callback_(std::move(out));
```
xianshijing-lk: nit, replace the whole block with 
if (callback_) {
  callback_(YuvFrame{std::move(frame), ts_us});
}

examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+void YuvSource::start() {
+  if (running_.exchange(true))
+    return;
+  thread_ = std::thread(&YuvSource::loop, this);
+}
+
+void YuvSource::stop() {
+  running_.store(false);
+  if (thread_.joinable())
+    thread_.join();
```
xianshijing-lk: I think .join() will hang the application if the recv() is blocking. Suggestion, set a recv timeout there

examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+      callback_(std::move(callback)) {
+  if (width_ > 0 && height_ > 0) {
+    frame_size_ = static_cast<std::size_t>(width_) *
+                  static_cast<std::size_t>(height_) * 3 / 2;
+  }
+}
+
+YuvSource::~YuvSource() { stop(); }
+
+void YuvSource::start() {
```
xianshijing-lk: I know it is an example, but wonder if you want the users to call start() / stop() multiple times ? 

or they are supposed to be called only once ? 

btw, if start() / stop() can be called multiple times WSAStartup(MAKEWORD(2, 2), &wsa) will likely won't work well.

examples/publish_yuv_source/yuv_source.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+} // namespace
+
+YuvSource::YuvSource(const std::string &host, std::uint16_t port, int width,
+                     int height, int fps, YuvFrameCallback callback)
+    : host_(host), port_(port), width_(width), height_(height), fps_(fps),
+      callback_(std::move(callback)) {
+  if (width_ > 0 && height_ > 0) {
+    frame_size_ = static_cast<std::size_t>(width_) *
```
xianshijing-lk: nit, do we need to validate the width and height to match the expected format

xianshijing-lk (commented):
thanks, it looks good, some comments.

Sorry for being late on reviewing, I was traveling international in the past week.

PR #66, opened 2026-02-26
DataTracks integration through FFI

stephen-derosa:
# Overview
Integrate DataTracks into the CPP SDK through the FFI

## Details
* setOnAudio/VideoFrameCallback() change signature to use track name
* setOnAudio/VideoFrameCallback() addition signature to use track name
* result.h file for floating up specific errors provided by the FFI
* update token gen script to run for all tokens needed
* Update client-sdk-rust submodule
* data_track_subscription: on pushFrame() called when rust sends new frame, assert(since that breaks cpp<->ffi<->rust agreement of rust handling buffering

## Testing
- src/tests/integration/test_data_track.cpp

## Examples
- [`examples/ping_pong`](https://github.com/livekit-examples/cpp-example-collection/pull/4/changes#diff-9c00330cf3fd9ea6030041f7efae9effe1a2acb068d2c9353ef1a0d0918afedd)

## Blocked By
[Rust SDKs PR](https://github.com/livekit/rust-sdks/pull/905)

Resolves BOT-268


include/livekit/data_track_frame.h
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2025 LiveKit
```
xianshijing-lk: nit, 2026?

include/livekit/data_track_info.h
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2025 LiveKit
```
xianshijing-lk: nit, 2026

include/livekit/local_data_track.h
```diff
@@ -0,0 +1,92 @@
...
+  ~LocalDataTrack() = default;
+
+  LocalDataTrack(const LocalDataTrack &) = delete;
+  LocalDataTrack &operator=(const LocalDataTrack &) = delete;
+
+  /// Metadata about this data track.
+  const DataTrackInfo &info() const noexcept { return info_; }
+
+  /// Raw FFI handle id for internal use.
+  uintptr_t ffi_handle_id() const noexcept { return handle_.get(); }
```
xianshijing-lk: why this needs to be public ?

include/livekit/local_participant.h
```diff
@@ -135,6 +142,20 @@ class LocalParticipant : public Participant {
...
+   * Data tracks carry arbitrary binary frames and are independent of the
+   * audio/video track hierarchy. The returned LocalDataTrack can push
+   * frames via tryPush() and be unpublished via unpublish().
+   *
+   * @param name  Unique track name visible to other participants.
+   * @return Shared pointer to the published data track.
+   * @throws std::runtime_error on FFI or publish failure.
+   */
+  std::shared_ptr<LocalDataTrack>
+  publishDataTrack(const std::string &name);
```
xianshijing-lk: why we need a publishDataTrack() here ? 

Why can't we re-use the same publishTrack and unpublishTrack ? 
Ideally, users will need to learn somethings specific for data track, it is like, if they know how to use audio track, then they know how to use video track, and data track.

.github/workflows/builds.yml
```diff
@@ -7,6 +7,7 @@ on:
       - src/**
       - include/**
       - examples/**
+      - bridge/**
```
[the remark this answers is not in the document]
stephen-derosa: sounds good -- i will make this a single `git mv` commit at the end so comments are easier to follow!

include/livekit/local_data_track.h
```diff
@@ -0,0 +1,92 @@
...
+  /// Whether the track is still published in the room.
+  bool isPublished() const;
+
+  /**
+   * Unpublish this data track from the room.
+   *
+   * After this call, tryPush() will fail and the track cannot be
+   * re-published. The underlying FFI handle is released.
+   */
+  void unpublish();
```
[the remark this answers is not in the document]
stephen-derosa: fair point, i had thought it was a little cleaner, but it doesnt follow the form of Audio/Video tracks. Updated.

ladvoc (commented):
Some initial comments, but I will make another pass. Looking clean so far!

include/livekit/data_track_subscription.h
```diff
@@ -0,0 +1,119 @@
...
+ *   while (sub->read(frame)) {
+ *     // process frame.payload
+ *   }
+ */
+class DataTrackSubscription {
+public:
+  struct Options {
+    /// Maximum buffered frames (count). 0 = unbounded.
+    /// When non-zero, behaves as a ring buffer (oldest dropped on overflow).
+    std::size_t capacity{0};
```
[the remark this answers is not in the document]
stephen-derosa: in general i lean towards putting things on the rust side for as you say uniform SDKs. But i dont there should be anything stopping client SDKs from implementing a capacity themselves. Plus in lieu of a rust configurable size this is a great stop gap.

include/livekit/data_track_subscription.h
```diff
@@ -0,0 +1,119 @@
...
+
+private:
+  friend class RemoteDataTrack;
+
+  DataTrackSubscription() = default;
+
+  void init(FfiHandle subscription_handle, const Options &options);
+
+  void onFfiEvent(const proto::FfiEvent &event);
+  void pushFrame(DataFrame &&frame);
```
ladvoc: issue: Data flow is unidirectional (i.e. publisher → subscriber), so there shouldn't be a push frame method for the subscription.

src/local_data_track.cpp
```diff
@@ -0,0 +1,65 @@
...
+  msg->set_track_handle(static_cast<uint64_t>(handle_.get()));
+  auto *pf = msg->mutable_frame();
+  pf->set_payload(frame.payload.data(), frame.payload.size());
+  if (frame.user_timestamp.has_value()) {
+    pf->set_user_timestamp(frame.user_timestamp.value());
+  }
+
+  proto::FfiResponse resp = FfiClient::instance().sendRequest(req);
+  const auto &r = resp.local_data_track_try_push();
+  return !r.has_error();
```
ladvoc: question: Should we be exposing the error reason to the caller? There are several reasons why pushing a frame could fail (e.g., the track is unpublished, buffer full, etc.). For the ROS bridge, a boolean might be fine if there is logging, but for the core SDK, I lean towards we should report the error reason.

Related question: the FFI interface currently flattens all errors to strings for simplicty, however, on the Rust side there are proper error enums that describe the reason for the error. Should I expose these over FFI rather than using strings? Can an enum describing the error reason map nicely into a C++ exception?

ladvoc (commented):
Some additional comments.

include/livekit/data_track_subscription.h
```diff
@@ -0,0 +1,119 @@
...
+class DataTrackSubscription {
+public:
+  struct Options {
+    /// Maximum buffered frames. 0 = unbounded.
+    /// When non-zero, behaves as a ring buffer (oldest dropped on overflow).
+    std::size_t capacity{0};
+  };
+
+  virtual ~DataTrackSubscription();
+
```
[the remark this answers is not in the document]
ladvoc: > At least one confusing thing about "subscription" is that this object can also call pushFrame(), so its not purely a subscription

From sync: data flow is unidirectional, so there should be no push method on the subscription.

include/livekit/data_track_subscription.h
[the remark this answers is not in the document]
ladvoc: Yes, planning to expose that in the FFI layer today.

bridge/src/livekit_bridge.cpp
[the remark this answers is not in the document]
stephen-derosa: done

xianshijing-lk (dismissed):

.github/workflows/builds.yml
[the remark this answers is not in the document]
stephen-derosa: no longer relevant with the bridge being deprecated :)

src/data_track_subscription.cpp
```diff
@@ -0,0 +1,180 @@
...
+void DataTrackSubscription::pushFrame(DataFrame &&frame) {
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  if (closed_ || eof_) {
+    return;
+  }
+
+  // rust side handles buffering, so we should only really ever have one item
+  if (queue_.size() >= 2) {
+    LK_LOG_ERROR("[DataTrackSubscription] Queue size is greater than 1, this "
```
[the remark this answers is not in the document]
stephen-derosa: done

examples/hello_livekit/hello_livekit.cpp
```diff
@@ -0,0 +1,230 @@
+/*
+ * Copyright 2025 LiveKit
```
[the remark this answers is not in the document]
stephen-derosa: done

examples/hello_livekit/hello_livekit.cpp
```diff
@@ -0,0 +1,230 @@
...
+  std::signal(SIGINT, handleSignal);
+#ifdef SIGTERM
+  std::signal(SIGTERM, handleSignal);
+#endif
+
+  livekit::initialize(livekit::LogLevel::Info, livekit::LogSink::kConsole);
+
+  // Shared: sender publishes its identity here so the receiver knows who
+  // to subscribe to (no hardcoded identity required).
+  std::atomic<std::string *> sender_identity{nullptr};
```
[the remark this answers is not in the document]
stephen-derosa: i broke this into two executables for easier readability/simpler

src/data_track_subscription.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+void DataTrackSubscription::pushFrame(DataFrame &&frame) {
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  if (closed_ || eof_) {
+    return;
+  }
+
+  // rust side handles buffering, so we should only really ever have one item
+  if (frame_.has_value()) {
```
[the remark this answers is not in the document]
stephen-derosa: done since if frame_ had a value, it would break the "rust handle buffers" contract we have.

include/livekit/data_track_frame.h
```diff
@@ -0,0 +1,45 @@
...
+  /** Arbitrary binary payload (the frame contents). */
+  std::vector<std::uint8_t> payload;
+
+  /**
+   * Optional application-defined timestamp.
+   *
+   * The proto field is a bare uint64 with no prescribed unit.
+   * By convention the SDK examples use microseconds since the Unix epoch.
+   */
+  std::optional<std::uint64_t> user_timestamp;
```
xianshijing-lk, line 46: nit, suggested adding things like 

DataFrame() = default;
DataFrame(DataFrame&&) = default;
DataFrame& operator=(DataFrame&&) = default;

These constructors will help compiler optimize those move, assignment use cases

src/local_data_track.cpp
xianshijing-lk: for SDKs used in real-time systems, in general, exceptions are discouraged for "expected" failures like "buffer full" or "track unpublished".

I think the best practice here might be returning an result, which will include the success or error codes/Enums ?

include/livekit/data_track_subscription.h
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * Typical usage:
+ *
+ *   auto sub = remoteDataTrack->subscribe();
+ *   DataFrame frame;
+ *   while (sub->read(frame)) {
+ *     // process frame.payload
+ *   }
+ */
+class DataTrackSubscription {
```
xianshijing-lk: maybe I asked this question before, I wonder if should be called data_track_stream ? 
or remote_data_track_stream ? 

so it will be aligned with audio_stream / video_stream

src/data_track_subscription.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+void DataTrackSubscription::onFfiEvent(const FfiEvent &event) {
+  if (event.message_case() != FfiEvent::kDataTrackSubscriptionEvent) {
+    return;
+  }
+
+  const auto &dts = event.data_track_subscription_event();
+  if (dts.subscription_handle() !=
+      static_cast<std::uint64_t>(subscription_handle_.get())) {
```
xianshijing-lk: this worries me a bit, onFfiEvent is a callback. If DataTrackSubscription is moved to a new memory location, the lambda captured this.

take a step back, why do we need to support the move operator ? do we see real use cases from it ?

src/data_track_subscription.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  // Signal the Rust side that we're ready to receive the next frame.
+  // The Rust SubscriptionTask uses a demand-driven protocol: it won't pull
+  // from the underlying stream until notified via this request.
+  proto::FfiRequest req;
+  auto *msg = req.mutable_data_track_subscription_read();
+  msg->set_subscription_handle(
+      static_cast<uint64_t>(subscription_handle_.get()));
+  FfiClient::instance().sendRequest(req);
```
xianshijing-lk: I think there are some races there. 
like one thread calls this read(), and pass this section under std::lock_guard<std::mutex> lock(mutex_); and get to this section.

meanwhile, close() is being called, and reset the subscription_handle_

src/data_track_subscription.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+  std::unique_lock<std::mutex> lock(mutex_);
+  cv_.wait(lock, [this] { return frame_.has_value() || eof_ || closed_; });
+
+  if (closed_ || (!frame_.has_value() && eof_)) {
+    return false;
+  }
+
+  out = std::move(frame_.value());
+  frame_.reset();
```
xianshijing-lk, line 66: nit, If the DataFrame have move constructor, you don't need this frame_.reset()

include/livekit/local_data_track.h
```diff
@@ -0,0 +1,112 @@
...
+   */
+  bool tryPush(const std::vector<std::uint8_t> &payload,
+               std::optional<std::uint64_t> user_timestamp = std::nullopt);
+  /**
+   * Try to push a frame to all subscribers of this track.
+   *
+   * @return true on success, false if the push failed (e.g. back-pressure
+   *         or the track has been unpublished).
+   */
+  bool tryPush(const std::uint8_t *data, std::size_t size,
```
xianshijing-lk: curiously, do we need all these 3 overloaded methods ? or we should push for one instead?

src/local_participant.cpp
```diff
@@ -286,6 +288,30 @@ LocalParticipant::PublicationMap LocalParticipant::trackPublications() const {
   return out;
 }
 
+std::shared_ptr<LocalDataTrack>
+LocalParticipant::publishDataTrack(const std::string &name) {
```
xianshijing-lk, line 292: curiously, what will happen if calling publishDataTrack with the same |name| more than once ?

src/local_data_track.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+bool LocalDataTrack::tryPush(const std::vector<std::uint8_t> &payload,
+                             std::optional<std::uint64_t> user_timestamp) {
+  DataFrame frame;
+  frame.payload = payload;
```
xianshijing-lk: this makes me wondering if we should allow developers to pass things like 
bool LocalDataTrack::tryPush(std::vector<std::uint8_t>&& payload,
                             std::optional<std::uint64_t> user_timestamp) { ... }

So that we can use the move constructor for the payload ?

src/local_participant.cpp
stephen-derosa: this will throw -- `publishDataTrack()` calls `FfiClient::publishDataTrackAsync` which boils into:
```
  proto::FfiRequest req;
  auto *msg = req.mutable_publish_data_track();
  msg->set_local_participant_handle(local_participant_handle);
  msg->mutable_options()->set_name(track_name);
  msg->set_request_async_id(async_id);

  try {
    proto::FfiResponse resp = sendRequest(req);
    if (!resp.has_publish_data_track()) {
      logAndThrow("FfiResponse missing publish_data_track");
    }
  } catch (...) {
    cancelPendingByAsyncId(async_id);
    throw;
  }
```

this matches the behavior of the `FfiClient::publishTrackAsync`

src/local_data_track.cpp
stephen-derosa: yeah great point -- much more efficient. I wonder if we even need:
```
bool LocalDataTrack::tryPush(const std::uint8_t *data, std::size_t size,
                             std::optional<std::uint64_t> user_timestamp) {
```
now, might be convenient for some users but it enables stuff to be slower and another step of Q: "hey why is this slow" A: "are you using the right tryPush() function?"

any reason not to keep the single more efficient tryPush()?

include/livekit/local_data_track.h
stephen-derosa: great point, and we have four `tryPush()` calls after adding
```
bool tryPush(std::vector<std::uint8_t> &&payload,
               std::optional<std::uint64_t> user_timestamp = std::nullopt);
```
I think this is too much. My opinion here would be to keep the efficient tryPush() call which uses std::move(), and keep:
```
bool tryPush(const DataFrame &frame);
```

I can forsee `bool tryPush(const DataFrame &frame);` being used to directly forward data from one room to another.

src/data_track_subscription.cpp
stephen-derosa: sweet! thank you!

include/livekit/data_track_frame.h
stephen-derosa: ah great point, updated!

src/data_track_subscription.cpp
stephen-derosa: This is a really good catch.

Since the public API provides a shared_ptr<DataTrackSubscription>, i dont think we do need to provide a move constructor. the pointer itself can be std::move(), not the object itself.

Side note:  I wonder how we can setup some tests to try to catch edge cases like this.

src/data_track_subscription.cpp
stephen-derosa: Nice thank you -- with the addition of this request call to the data tracks API i missed putting this in the lock!

include/livekit/local_data_track.h
ladvoc: I agree, we should anticipate data track frames having more metadata fields in the future, and we likely won't want overloads for all of them. Also, small nitpick: in the other SDKs `DataFrame` is called `DataTrackFrame`.

src/data_track_subscription.cpp
stephen-derosa: I also added better logic to close() which:
- gets the lock
- sets closed_ = true
- resets the subscription handle
- sets listener_id_ = 0
- releases the lock
- calls ffi RemoveListener()

src/local_data_track.cpp
stephen-derosa: thanks for the guidance. I like the enum idea. I will implement something like:
```
Enum ErrorCodes
{
  QUEUE_FULL,
  ...,
  UNKOWN
}
```

I will also adopt this going forward when applicable!

include/livekit/data_track_subscription.h
stephen-derosa: i had originally named as data_track_subscription because:
a. data tracks follow a slightly different format than video/audio
b. conceptually it was easier for me to understand.

However, i think your point is especially valid because
a. only data track _publish_ and _unpublish_ are different
b. consistency is important. 

Will update once all other comments are resolved to avoid confusing referencing of MR comments!

src/local_data_track.cpp
ladvoc: Note: on the Rust side, all fallible operations define error enums which enumerate all possible reasons for the failure. Currently, these are converted to strings when passed over FFI, but we can expose the actual enums if that's helpful.

include/livekit/data_track_subscription.h
ladvoc: I would push for keeping it as subscription since there are semantic differences between the two:
- _AudioStream_ and _VideoStream_ for are purely local frame taps. Dropping one removes a local sink, but the SFU subscription is unaffected.
- _DataTrackSubscription_ controls the subscription itself; the first _subscribe()_ call triggers SFU signaling, and dropping the last subscription for a track unsubscribes from the SFU. It's not a passive reader, but rather managing a resource.

src/local_data_track.cpp
stephen-derosa: i went here with returning a templated `Result` which has the form:
```
template <typename T, typename E> class Result
```

An example is `<std::shared_ptr<DataTrackSubscription>, DataTrackErrorCode>` . I think this is a _little_ complex for the current state but it should scale really well for new error codes! LMK what you think, we can simplify if need be.

stephen-derosa: @ladvoc we can definelty support the enums! As of now we are kind of bastardizing them by setting them custom based on some checks. For example stuff like:
```
  if (!handle_.valid()) {
    return Result<void, DataTrackError>::failure(
        DataTrackError{DataTrackErrorCode::INVALID_HANDLE,
                       "LocalDataTrack::tryPush: invalid FFI handle", false});
  }
```

src/local_data_track.cpp
xianshijing-lk: I would think we will just keep the single more efficient tryPush() so that developers won't pay for the cost without knowing those might involve extra memcopy under the hood. Those performance might be a bottleneck if they are sending high frequency large data.

src/data_track_subscription.cpp
xianshijing-lk: I guess as long as you release the log in the middle of the read function, and close() function reset the subscription_handle_, you will still run into the race here where the invalid subscription_handle_ is being used

xianshijing-lk: maybe it is better to double check it with the AI ?

include/livekit/local_data_track.h
stephen-derosa: will rename this to DataTrackFrame to align with other SDKs once other comments are resolved for easier tracing.

src/local_data_track.cpp
stephen-derosa: duplicate thread of here:
https://github.com/livekit/client-sdk-cpp/pull/66#discussion_r3002369500

include/livekit/data_track_subscription.h
stephen-derosa: if the trade off is align with cpp or align with other SDKs, aligning with other SDKs is more future proofing.

src/data_track_subscription.cpp
stephen-derosa: i think the race is gated by checks of `if (closed_) return;` before subscription_handle_ is used (except for init).

I did ask Codex GPT 5.4 though and it said:
```
In the current read() path, the handle is read and used while mutex_ is still held in src/data_track_subscription.cpp (line 39). Specifically, read():

- takes mutex_ at src/data_track_subscription.cpp (line 41)
- checks closed_ / eof_
- copies subscription_handle_.get() to a local at src/data_track_subscription.cpp (line 49)
- sends the FFI read request at src/data_track_subscription.cpp (line 58)
```

include/livekit/data_frame.h
```diff
@@ -0,0 +1,60 @@
...
+  DataFrame(const DataFrame&) = default;
+  DataFrame(DataFrame&&) noexcept = default;
+  DataFrame& operator=(const DataFrame&) = default;
+  DataFrame& operator=(DataFrame&&) noexcept = default;
+
+  explicit DataFrame(std::vector<std::uint8_t>&& p,
+                     std::optional<std::uint64_t> ts = std::nullopt) noexcept
+      : payload(std::move(p)), user_timestamp(ts) {}
+
+  static DataFrame fromOwnedInfo(const proto::DataTrackFrame &owned);
```
xianshijing-lk: nit, add a comment that this fromOwnedInfo() is used internally only, not public API.

include/livekit/data_track_subscription.h
xianshijing-lk: @ladvoc , can you elaborate a bit more ? 
From what I am seeing DataTrackSubscription class does not have subscribe() function.
It is created with auto sub_result = remoteDataTrack->subscribe();
From what I am seeing, developers can subscript to the track multiple times, similar to audioStream / videoStream.

I don't think DataTrackSubscription control the subscription, but the DataTrack does.

xianshijing-lk:
taking a final look

src/local_data_track.cpp
stephen-derosa: this has been completed.

include/livekit/data_track_subscription.h
stephen-derosa: based on this [thread](https://live-kit.slack.com/archives/C09G6SCR620/p1775000773631949?thread_ts=1774901967.943499&cid=C09G6SCR620) i renamed `data_track_subscription` -> `data_track_stream`

include/livekit/local_data_track.h
stephen-derosa: done

include/livekit/data_track_subscription.h
xianshijing-lk: Thanks

xianshijing-lk (approved):

PR #78, opened 2026-03-20
Deprecate BridgeVideoTrack and BridgeAudioTrack

stephen-derosa:
## LocalParticipant
- moves publications to be owned by the track
- overloaded local_participant function to createDataTrack
- helper function to create and publish track

## Motivation
Maintaining the bridge and the base SDK is highly undesirable. This exemplifies a way to at the very least heavily reduce redundant-ish code.


include/livekit/local_participant.h
```diff
@@ -55,6 +57,7 @@ class LocalParticipant : public Participant {
 public:
   using PublicationMap =
       std::unordered_map<std::string, std::shared_ptr<LocalTrackPublication>>;
+  using TrackMap = std::unordered_map<std::string, std::weak_ptr<Track>>;
```
stephen-derosa: note: we use weak_ptr here because the local_participant doesnt _own_ the tracks, the application does. However, the local_participant needs to call setPublication(nullptr) when a track is unpublished.

include/livekit/local_audio_track.h
```diff
@@ -61,6 +63,22 @@ class LocalAudioTrack : public Track {
   createLocalAudioTrack(const std::string &name,
                         const std::shared_ptr<AudioSource> &source);
 
+  /// Creates a local audio track with a new \ref AudioSource for the given
+  /// sample rate and channel count (see \ref audioSource).
+  static std::shared_ptr<LocalAudioTrack>
+  createLocalAudioTrack(const std::string &name, int sample_rate,
+                        int num_channels, TrackSource track_source);
```
xianshijing-lk: TrackSource is an enum, right? 

C++ does not have media handling today, what is the point to specify a TrackSource ?

stephen-derosa:
closed because its included here: https://github.com/livekit/client-sdk-cpp/pull/79

PR #83, opened 2026-04-02
Move all examples to cpp-example-collection

stephen-derosa:
Move all examples to cpp-example-collection repo.

Depends on:
https://github.com/livekit-examples/cpp-example-collection/pull/4


CMakeLists.txt
```diff
@@ -690,7 +690,8 @@ install(FILES
 add_subdirectory(bridge)
 
 if(LIVEKIT_BUILD_EXAMPLES)
-  add_subdirectory(examples)
+  include(cpp-example-collection)
+  livekit_configure_cpp_example_collection()
 endif()
```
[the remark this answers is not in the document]
stephen-derosa: updated comments in previous commit

cmake/spdlog.cmake
```diff
@@ -71,7 +71,7 @@ FetchContent_Declare(
 set(SPDLOG_BUILD_SHARED OFF CACHE BOOL "" FORCE)
 set(SPDLOG_BUILD_EXAMPLE OFF CACHE BOOL "" FORCE)
 set(SPDLOG_BUILD_TESTS OFF CACHE BOOL "" FORCE)
-set(SPDLOG_INSTALL OFF CACHE BOOL "" FORCE)
+set(SPDLOG_INSTALL ON CACHE BOOL "" FORCE)
```
xianshijing-lk: why do we need to install SPDLOG here

cmake/cpp-example-collection.cmake
```diff
@@ -0,0 +1,70 @@
...
+  set(LIVEKIT_CPP_EXAMPLES_INSTALL_PREFIX "${_lk_examples_install_default}"
+    CACHE PATH "Install prefix used by cpp-example-collection")
+
+  get_filename_component(_lk_examples_binary_default
+    "${CMAKE_BINARY_DIR}/cpp-example-collection-build" ABSOLUTE)
+  set(LIVEKIT_CPP_EXAMPLES_BINARY_DIR "${_lk_examples_binary_default}"
+    CACHE PATH "Build directory for cpp-example-collection")
+
+  get_filename_component(LIVEKIT_CPP_EXAMPLES_LIVEKIT_DIR
+    "${LIVEKIT_CPP_EXAMPLES_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/cmake/LiveKit"
```
xianshijing-lk, line 25: curiously, where CMAKE_INSTALL_LIBDIR is defined ? and what is its value ?

cmake/cpp-example-collection.cmake
```diff
@@ -0,0 +1,70 @@
...
+
+  add_custom_target(cpp_example_collection ALL
+    COMMAND ${CMAKE_COMMAND} --install "${CMAKE_BINARY_DIR}"
+            --prefix "${LIVEKIT_CPP_EXAMPLES_INSTALL_PREFIX}"
+    COMMAND ${CMAKE_COMMAND} -S "${LIVEKIT_CPP_EXAMPLES_SOURCE_DIR}"
+            -B "${LIVEKIT_CPP_EXAMPLES_BINARY_DIR}"
+            ${_lk_examples_configure_args}
+    COMMAND ${CMAKE_COMMAND} --build "${LIVEKIT_CPP_EXAMPLES_BINARY_DIR}"
+            --config "$<CONFIG>"
+    DEPENDS install_livekit_sdk_for_examples
```
xianshijing-lk, line 64: this seems to install it twice ?

xianshijing-lk (approved):

cmake/spdlog.cmake
stephen-derosa: thanks this got caught in a rebase

cmake/cpp-example-collection.cmake
stephen-derosa: fixed!

cmake/cpp-example-collection.cmake
stephen-derosa: CMAKE_INSTALL_LIBDIR is defined by CMake's built-in GNUInstallDirs module. Its value is typically lib, but can be lib64 on some 64-bit Linux distros. Claude pointed out:
"
OK so GNUInstallDirs is included at line 551 and the function is called at line 693, so ordering is fine — CMAKE_INSTALL_LIBDIR will be defined. But it's still an implicit dependency that could break if someone reorders things.

It's cleaner to make the dependency explicit by including GNUInstallDirs inside the function as a guard. include_guard doesn't apply here since GNUInstallDirs is idempotent, but calling include() again is harmless and makes the function self-contained.
"

I added the include to this cmake file

.github/workflows/builds.yml
stephen-derosa, line 1: @alan-george-lk FYI ths MR locks our actions to a specific hash and checks licenses. (i guess there was a malware attack through un-pinned actions to run python on a open source project's runner.)

.github/workflows/publish-docs.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         # Note: submodules and LFS not needed for docs generation
         # Doxygen only reads from include/, docs/, README.md, and examples/
```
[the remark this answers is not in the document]
stephen-derosa: not relevant to this body of work

PR #87, opened 2026-04-08
Support for compile_commands.json/clangd

alan-george-lk:
This PR adds support for `compile_commands.json` (Mac/Linux). This is implemented using CMake's built-in feature, but additionally the build script symlinks the file to the root of the directory (but is `.gitignore`'d). 

This enables a few features:
- Intellisense autocomplete via VSCode/Cursor extension `llvm-vs-code-extensions.vscode-clangd`
- `clang-tidy` support in a future PR for static analysis checks on the repository

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -152,6 +152,11 @@ configure() {
       cmake -S . -B "${BUILD_DIR}" -DCMAKE_BUILD_TYPE="${BUILD_TYPE}"
     fi
   fi
+
+  if [[ -f "${BUILD_DIR}/compile_commands.json" ]]; then
+    ln -sf "${BUILD_DIR}/compile_commands.json" "${PROJECT_ROOT}/compile_commands.json"
+    echo "==> Symlinked compile_commands.json -> ${BUILD_DIR}/compile_commands.json"
+  fi
 }
 
 build() {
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -152,6 +152,11 @@ configure() {
       cmake -S . -B "${BUILD_DIR}" -DCMAKE_BUILD_TYPE="${BUILD_TYPE}"
     fi
   fi
+
+  if [[ -f "${BUILD_DIR}/compile_commands.json" ]]; then
```
stephen-derosa, line 156: is the file compile_commands.json generated at build time?

alan-george-lk: Yup, see: https://github.com/livekit/client-sdk-cpp/blob/add1f5f4f7a1ba22d65947205057272ca32fbb11/CMakePresets.json#L42-L56

alan-george-lk: Lines `42` and `56` for example

xianshijing-lk, line 156: how about the windows support ? which is at build.cmd

xianshijing-lk (approved):

build.sh
alan-george-lk: Windows/Visual studio will be handled a bit differently, it would take a big refactor of the Windows build pipeline to handle that in this case. Will reserve that decision for future user need.

stephen-derosa (approved):

PR #89, opened 2026-04-09
Running unit tests in CI

alan-george-lk:
This PR adjusts the CI to do the following:

- Build stage now builds `release-tests` and `release-examples`
- An additional step to run the unit tests is added
- An existing folder `integration/` existed in the tests directory, but contained unit-level testing. A new folder `unit/` was added, which is now being executed in CI
  - Note: "Integration" tests are scoped as tests that require a LiveKit server instance running, and ideally will be invoked in CI on a later PR to match other SDK convention
- Changes needed to get the unit tests buildable across all platforms

Note: part of the time spent on this PR was spent investigating a separate testing stage for unit testing (and upcoming `clang-tidy`), but passing artifacts between stages became pretty cumbersome and affected overall build usefulness. I think a future refactor of the CI jobs should be considered to improve this.


.github/workflows/builds.yml
```diff
@@ -61,23 +61,23 @@ jobs:
         include:
           - os: ubuntu-latest
             name: linux-x64
-            build_cmd: ./build.sh release-examples
+            build_cmd: ./build.sh release-tests
```
stephen-derosa: should we be using build-all here? note it will add significant time since it builds debug/release tests/exampls

.github/workflows/builds.yml
```diff
@@ -181,105 +181,28 @@ jobs:
...
-              return 0
-            fi
-            p="${examples_base}/${dir}/Release/${name}"
-            if [[ -x "${p}" ]]; then
-              printf '%s' "${p}"
-              return 0
-            fi
-            return 1
-          }
-          for pair in "SimpleRoom:simple_room" "SimpleRpc:simple_rpc" "SimpleDataStream:simple_data_stream"; do
```
stephen-derosa, line 210: is it safe to completely remove the smoke tests? Does it make sense to keeo them to ensure that they are compiled and be executed?

.github/workflows/builds.yml
```diff
@@ -181,105 +181,28 @@ jobs:
...
-            } else {
-              Write-Host "ERROR: ${name} not found under ${examplesBase}/${dir}/"
-              $failed = $true
-            }
-          }
-          if ($failed) { exit 1 } else { exit 0 }
+          & "${{ matrix.build_dir }}/bin/livekit_unit_tests.exe" `
+            --gtest_output=xml:${{ matrix.build_dir }}/unit-test-results.xml
+
+      - name: Upload test results
```
stephen-derosa, line 304: <3

src/tests/unit/test_audio_processing_module.cpp
```diff
@@ -801,6 +801,7 @@ TEST_F(AudioProcessingModuleTest, AGCAttenuatesLoudSpeech) {
   int sample_rate = 0;
   int num_channels = 0;
 
+  // TODO: figure out what generates this welcome.wav file such that this test isn't skipped
```
stephen-derosa: please create a ticket for this and use TODO(project-ticketid). I am guilt of not doing this as much as i should.

alan-george-lk: @xianshijing-lk insights here?

stephen-derosa (approved):
Looks good to me thanks for doing this!

src/tests/unit/test_audio_processing_module.cpp
alan-george-lk: This was a TODO for me before this PR, ideally I'll fix it before we merge, otherwise 100% will do

.github/workflows/builds.yml
alan-george-lk: The smoke tests were failing to build early in this endeavor, that might have been before I split the unit tests out. Can put this back, let me review and get back to you

.github/workflows/builds.yml
alan-george-lk: Open to it, I can profile that. Would be good to build all but yeah, will increase time

.github/workflows/builds.yml
```diff
@@ -61,23 +61,23 @@ jobs:
...
             build_dir: build-release
           - os: ubuntu-24.04-arm
             name: linux-arm64
-            build_cmd: ./build.sh release-examples
+            build_cmd: ./build.sh release-tests
             build_dir: build-release
           - os: macos-latest
             name: macos-arm64
-            build_cmd: ./build.sh release-examples
+            build_cmd: ./build.sh release-tests
```
xianshijing-lk: curiously, should we build all instead ? 

like
./build.sh release-all

.github/workflows/builds.yml
```diff
@@ -181,105 +181,28 @@ jobs:
         shell: pwsh
         run: ${{ matrix.build_cmd }}
 
-      # ---------- Smoke test cpp-example-collection binaries ----------
```
xianshijing-lk, line 186: I would assume it is still useful to run this cpp-example-collection ? 

If any change is required inside the cpp-example-collection, that raise a flag that we might have a breaking API change ?

src/tests/CMakeLists.txt
```diff
@@ -24,6 +24,91 @@ FetchContent_MakeAvailable(googletest)
 enable_testing()
 include(GoogleTest)
 
+# ============================================================================
+# Unit Tests
+# ============================================================================
+
+file(GLOB UNIT_TEST_SOURCES
+  "${CMAKE_CURRENT_SOURCE_DIR}/unit/*.cpp"
```
xianshijing-lk, line 32: nit, generally it is better to add all the cpp files explicitly to a list.

For instance, with this current code, when adding new test files, CMake won't automatically rerun.

src/tests/CMakeLists.txt
```diff
@@ -24,6 +24,91 @@ FetchContent_MakeAvailable(googletest)
...
+
+  target_include_directories(livekit_unit_tests
+    PRIVATE
+      ${LIVEKIT_ROOT_DIR}/include
+      ${LIVEKIT_ROOT_DIR}/src
+      ${LIVEKIT_BINARY_DIR}/generated
+      ${Protobuf_INCLUDE_DIRS}
+  )
+  if(TARGET absl::base)
+    get_target_property(_livekit_unit_test_absl_inc absl::base INTERFACE_INCLUDE_DIRECTORIES)
```
xianshijing-lk: This smells a bit. 

Why do we need such special handling of absl::base ? 

I wonder if we have some real problem with our SDK, technically, consumers of livekit SDK needs to manually have the include dirs from absl::base unless the tests use them ?

src/tests/CMakeLists.txt
```diff
@@ -24,6 +24,91 @@ FetchContent_MakeAvailable(googletest)
...
+      GTest::gtest_main
+      GTest::gmock
+  )
+
+  target_include_directories(livekit_unit_tests
+    PRIVATE
+      ${LIVEKIT_ROOT_DIR}/include
+      ${LIVEKIT_ROOT_DIR}/src
+      ${LIVEKIT_BINARY_DIR}/generated
+      ${Protobuf_INCLUDE_DIRS}
```
xianshijing-lk: why protobuf is needed ? 

maybe some tests are not correctly written ?

src/tests/unit/test_audio_processing_module.cpp
xianshijing-lk: It was something very basic, I generated it by using the the say cmd on macbook, like
say "Hello world" -o hello_world.wav

alan-george-lk: @xianshijing-lk Do we want this included in the unit tests/committed to the repo via Git LFS?

.github/workflows/builds.yml
alan-george-lk: Not presently an option in `./build.sh`, I am testing running `release-tests` and `release-examples` sequentially

.github/workflows/builds.yml
alan-george-lk: Per discussions modified to build both `release-tests` && `release-examples` targets of the build scripts

.github/workflows/builds.yml
alan-george-lk: Smoke tests restored per this change: https://github.com/livekit/client-sdk-cpp/pull/89#discussion_r3074769680

.github/workflows/builds.yml
alan-george-lk: Resolved per: https://github.com/livekit/client-sdk-cpp/pull/89#discussion_r3074769680

.github/workflows/builds.yml
alan-george-lk: Restored this stage after: https://github.com/livekit/client-sdk-cpp/pull/89#discussion_r3074769680

src/tests/CMakeLists.txt
alan-george-lk: I believe this change was needed for the Windows stage to succeed, let me review this again and will reply back here with removal or justification.

src/tests/CMakeLists.txt
alan-george-lk: This may have been a copy/paste mistake, reviewing now

src/tests/unit/test_audio_processing_module.cpp
alan-george-lk: Resolved via `git-lfs` pull in CI step

src/tests/CMakeLists.txt
alan-george-lk: @xianshijing-lk I went this route because the existing tests did it: https://github.com/livekit/client-sdk-cpp/blob/main/src/tests/CMakeLists.txt#L31-L34

I have resolved this PR to individually list out the unit test sources.

However, I think we should be consistent. Would you like me to make the other test types explicit in the sources they bring in?

src/tests/CMakeLists.txt
alan-george-lk: This seems to be copy/pasted from the existing `integration` test section: https://github.com/livekit/client-sdk-cpp/blob/main/src/tests/CMakeLists.txt#L49-L55

Removed

src/tests/CMakeLists.txt
alan-george-lk: Removed. This seemed to be a copy/paste from the `integration` tests section: https://github.com/livekit/client-sdk-cpp/blob/main/src/tests/CMakeLists.txt#L56-L63

.github/workflows/docker-images.yml
```diff
@@ -0,0 +1,363 @@
...
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  detect-changes:
+    name: Detect Docker image changes
+    runs-on: ubuntu-latest
+    outputs:
+      base_changed: ${{ steps.changes.outputs.base_changed }}
```
xianshijing-lk, line 30: I wonder if it will be easier if we output things like 
outputs:
  base_tag_amd64: ${{ steps.refs.outputs.base_tag_amd64 }}
  base_tag_arm64: ${{ steps.refs.outputs.base_tag_arm64 }}


where will be computed in one place like 
if [[ "${base_changed}" == "true" ]]; then
  echo "base_tag_amd64=base-${base_hash}-amd64" >> "$GITHUB_OUTPUT"
  echo "base_tag_arm64=base-${base_hash}-arm64" >> "$GITHUB_OUTPUT"
else
  echo "base_tag_amd64=base-main-amd64" >> "$GITHUB_OUTPUT"
  echo "base_tag_arm64=base-main-arm64" >> "$GITHUB_OUTPUT"
fi


then we can consume the outputs like 
--build-arg BASE_IMAGE="${{ needs.detect-changes.outputs.base_image }}:${{ needs.detect-changes.outputs.base_tag_amd64 }}"

Similarly for arm64

xianshijing-lk (approved):
one nit, lgtm if the docker builds are functional and fast.

src/tests/CMakeLists.txt
xianshijing-lk: It is a nit pick, so it is definitely fine to keep using the GLOB. For SDK source, I will definitely go for an explicit list. For tests, we can stick with GLOB,

.github/workflows/docker-images.yml
alan-george-lk: This change wasn't intended for this PR, after merging `main` in it is gone

.github/workflows/builds.yml
```diff
@@ -90,7 +90,10 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           submodules: recursive
-          fetch-depth: 0
+          fetch-depth: 1
+
+      - name: Pull LFS files
+        run: git lfs pull
```
stephen-derosa, line 96: guess git lfs is installed on the runners by default?

stephen-derosa (approved):
lgtm -- thanks for setting this up!

.github/workflows/builds.yml
alan-george-lk: Appears so

PR #91, opened 2026-04-09
AGENTS.md

stephen-derosa:
# Overview
An AGENTS.MD file for guiding cursor on development inside the project

Key points
- Architecture
- Build preferences
- Testing
- Formatting
- comments on readability vs performance

Any other points?

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -145,7 +145,7 @@ When adding new client facing functionality, add a new test case to the existing
 When adding new client facing functionality, add benchmarking to understand the limitations of the new functionality.
 
 ## Formatting
-Adhere to the formatting rules if TODO: @alan
+Use clang formatting that aligns with the existing codebase.
 
 ## Deprecated / Out of Scope
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,158 @@
+# AGENTS.md — LiveKit C++ Client SDK
+
+## Project Overview
+
+This is **client-sdk-cpp**, the official LiveKit C++ client SDK. It wraps a Rust core (`client-sdk-rust/`) via a protobuf-based FFI bridge. All WebRTC, networking, and media logic lives in Rust; the C++ layer provides an ergonomic API for C++ consumers.
+
+## Architecture
+
+### Core Principle
+Rust owns as much of the business logic as possible. If a feature may be used by another SDK it should be implemented in Rust. Since this is an SDK,
+ensure backwards compatibility is maintained when possible.
+
+
+### Platform Support
+The SDK must be supported on the following platforms:
+1. windows x64
+2. linux x86_64
+3. linux arm64
+3. macOS x86_64
+4. macOS arm64
+
+
+SDK features follow pattern:
+
+1. Serialize a protobuf `FfiRequest` message.
+2. Send it to Rust via `FfiClient::instance().sendRequest(req)`.
+3. Receive a synchronous `FfiResponse` or an asynchronous `FfiEvent` callback.
+4. Expose the result through the C++ API.
+
+When making larger scale changes, check with the developer before committing to architecture changes involving changes to the `client-sdk-rust/` submodule.
+
+### Directory Layout
+
+| Path | Description |
+|------|-------------|
+| `include/livekit/` | Public API headers (what SDK consumers include) |
+| `src/` | Implementation files and internal-only headers (`ffi_client.h`, `lk_log.h`, etc.) |
+| `src/tests/` | Google Test integration and stress tests |
+| `examples/` | In-tree example applications |
+| `bridge/` | **Deprecated** C-style bridge layer — do not add new functionality |
+| `client-sdk-rust/` | Git submodule holding the Rust core of the SDK|
+| `client-sdk-rust/livekit-ffi/protocol/*.proto` | FFI contract (protobuf definitions, read-only reference) |
+| `cmake/` | Build helpers (`protobuf.cmake`, `spdlog.cmake`, `LiveKitConfig.cmake.in`) |
+| `docker/` | Dockerfile for CI and SDK distribution images |
+| `.github/workflows/` | GitHub Actions CI workflows |
+
+### Key Types
+
+- **`FfiClient`** — Singleton that sends FFI requests to Rust and dispatches event callbacks. Defined in `src/ffi_client.h` (internal).
+- **`FfiHandle`** — RAII wrapper for Rust-owned resource handles; drop releases the FFI resource.
+- **`Room`** — Central object for connecting to a LiveKit server room.
+- **`RoomDelegate`** — Virtual callback interface for room lifecycle events.
+- **`SubscriptionThreadDispatcher`** — Owns callback registrations and per-subscription reader threads for audio, video, and data tracks. `Room` delegates callback management here.
+- **`LocalParticipant` / `RemoteParticipant`** — Participant objects for publishing and receiving tracks.
+
+## Build
+for building, use the build.sh script for Linux and macOS, and the build.cmd script for Windows. Do not invoke CMake directly.
+
+```
+./build.sh debug              # Debug build
+./build.sh debug-tests        # Debug build with tests
+./build.sh debug-examples     # Debug build with examples
+./build.sh release            # Release build
+./build.sh release-tests      # Release build with tests
+./build.sh release-examples   # Release build with examples
+./build.sh build-all          # All of the above
+./build.sh clean              # Clean build artifacts
+./build.sh clean-all          # Full clean (C++ + Rust targets)
+```
+
+**Requirements:** CMake 3.20+, C++17, Rust toolchain (cargo), protoc. On macOS: `brew install cmake ninja protobuf abseil spdlog`. On Linux: see the CI workflow for apt packages.
+
+### SDK Packaging
+
+```
+./build.sh release --bundle --prefix /path/to/install
+```
+
+This installs headers, libraries, and CMake config files so downstream projects can use `find_package(LiveKit CONFIG)`.
+
+## Coding Conventions
+
+### Copyright Header
+
+All source files must have the LiveKit Apache 2.0 copyright header. Use the current year for new files. Do not copyright externally pulled/cloned files.
+
+### Logging
+
+- Use `LK_LOG_*` macros from `src/lk_log.h` (internal header, **not** public API).
+- `LK_LOG_ERROR` — when something is broken.
+- `LK_LOG_WARN` — when something is unexpected but recoverable.
+- `LK_LOG_INFO` — initialization, critical state changes.
+- `LK_LOG_DEBUG` — potentially useful diagnostic info.
+- Do not spam logs. Keep it concise.
+- In production/internal SDK code, do not use `std::cout`, `std::cerr`, `printf`, or raw spdlog calls; use the logging APIs instead. Tests may use standard streams sparingly for ad hoc diagnostics when appropriate.
+- The public logging API is `include/livekit/logging.h` (`setLogLevel`, `setLogCallback`). spdlog is a **private** dependency — it must never leak into public headers.
+
+### Public API Surface
+
+- Keep public APIs small. Minimize what goes into `include/livekit/`.
+- Never introduce backwards compatibility breaking changes to the public API.
+- `lk_log.h` lives under `src/` (internal). The public-facing logging API is `include/livekit/logging.h`.
+- spdlog must not appear in any public header or installed header.
+
+### Error Handling
+
+- Throw exceptions for broken/unrecoverable states.
+- Use `LK_LOG_WARN` for non-fatal unexpected conditions.
+- Use `Result<T, E>` for operations that can fail with typed errors (e.g., data track publish/subscribe).
+
+### Git Practices
+
+- Use `git mv` when moving or renaming files.
+
+### CMake
+
+- spdlog is linked **PRIVATE** to the `livekit` target. It must not appear in exported/installed dependencies.
+- protobuf is vendored via FetchContent on non-Windows platforms; Windows uses vcpkg.
+- The CMake install produces a `find_package(LiveKit CONFIG)`-compatible package with `LiveKitConfig.cmake`, `LiveKitTargets.cmake`, and `LiveKitConfigVersion.cmake`.
+
+### Readability and Performance
+Code should be easy to read and understand. If a sacrifice is made for performance or readability, it should be documented.
+## Dependencies
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+cd build-debug && ctest
+```
+
+Integration tests (`src/tests/integration/`) cover: room connections, callbacks, data tracks, RPC, logging, audio processing, and the subscription thread dispatcher.
+
+When adding new client facing functionality, add a new test case to the existing test suite.
+When adding new client facing functionality, add benchmarking to understand the limiations of the new functionality.
+
+## Formatting
+Adhere to the formatting rules if TODO: @alan
```
stephen-derosa: @alan-george-lk i wonder what the best way t provide this is. My reading on [Agents.md](https://agents.md/) leads me to believe we should just put code snippets directly inline. Does clang-tidy provide a list or would it be reasonable to ask cursor:
"Write me an AGENTS.md ## Formatting section which has code examples to be followed for clang-tidy cpp formatting" (or a better prompt)

AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+- **`FfiHandle`** — RAII wrapper for Rust-owned resource handles; drop releases the FFI resource.
+- **`Room`** — Central object for connecting to a LiveKit server room.
+- **`RoomDelegate`** — Virtual callback interface for room lifecycle events.
+- **`SubscriptionThreadDispatcher`** — Owns callback registrations and per-subscription reader threads for audio, video, and data tracks. `Room` delegates callback management here.
+- **`LocalParticipant` / `RemoteParticipant`** — Participant objects for publishing and receiving tracks.
+
+## Build
+
+Use `./build.sh` — always. Do not invoke CMake directly.
+
```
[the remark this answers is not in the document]
stephen-derosa: @copilot apply changes based on [this feedback](https://github.com/livekit/client-sdk-cpp/pull/91#discussion_r3060455647)

AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+| Path | Description |
+|------|-------------|
+| `include/livekit/` | Public API headers (what SDK consumers include) |
+| `src/` | Implementation files and internal-only headers (`ffi_client.h`, `lk_log.h`, etc.) |
+| `src/tests/` | Google Test integration and stress tests |
+| `examples/` | In-tree example applications |
+| `bridge/` | **Deprecated** C-style bridge layer — do not add new functionality |
+| `client-sdk-rust/` | Git submodule — **read-only, never modify** |
+| `client-sdk-rust/livekit-ffi/protocol/*.proto` | FFI contract (protobuf definitions, read-only reference) |
```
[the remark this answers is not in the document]
stephen-derosa: @copilot apply changes based on [this feedback](https://github.com/livekit/client-sdk-cpp/pull/91#discussion_r3060455715)

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+Rust owns as much of the business logic as possible. If a feature may be used by another SDK it should be implemented in Rust. Since this is an SDK,
+ensure backwards compatibility is maintained when possible.
+
+
+### Platform Support
+The SDK must be supported on the following platforms:
+1. windows x64
+2. linux x86_64
+3. linux arm64
+3. macOS x86_64
```
xianshijing-lk: there are two 3 here

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+4. macOS arm64
+
+
+SDK features follow pattern:
+
+1. Serialize a protobuf `FfiRequest` message.
+2. Send it to Rust via `FfiClient::instance().sendRequest(req)`.
```
xianshijing-lk, line 26: there are two types of ffi messages here, one is synchronous call that we send a ffi request to rust ffi layer, and get back the FfiResponse. 

Another one is async calls:
c++ setup a async handler that listens to event with a |request_async_id|
Serialize a protobuf `FfiRequest` message with a |request_async_id|
Send it to Rust via `FfiClient::instance().sendRequest(req)`.
Receive a synchronous `FfiResponse` and an asynchronous `FfiEvent` callback.

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+| `src/` | Implementation files and internal-only headers (`ffi_client.h`, `lk_log.h`, etc.) |
+| `src/tests/` | Google Test integration and stress tests |
+| `examples/` | In-tree example applications |
+| `bridge/` | **Deprecated** C-style bridge layer — do not add new functionality |
+| `client-sdk-rust/` | Git submodule holding the Rust core of the SDK|
+| `client-sdk-rust/livekit-ffi/protocol/*.proto` | FFI contract (protobuf definitions, read-only reference) |
+| `cmake/` | Build helpers (`protobuf.cmake`, `spdlog.cmake`, `LiveKitConfig.cmake.in`) |
+| `docker/` | Dockerfile for CI and SDK distribution images |
+| `.github/workflows/` | GitHub Actions CI workflows |
+
```
xianshijing-lk, line 83: I’m wondering if we should add a dedicated section to clarify the FFI design rules. For example, we should explicitly state that the main implementation logic lives in Rust, and that the C++ layer must remain a thin wrapper over the FFI.

It would also be helpful to include a section describing the threading model, covering questions like:

Is FfiClient thread-safe? (this should be clearly defined)
On which threads are RoomDelegate callbacks invoked?
Are they coming from FFI/Tokio threads?
In the FFI layer, are audio callbacks executed on a dedicated high-priority Tokio runtime?
What are the threading models for:
LocalAudioTrack / LocalVideoTrack
AudioStream / VideoStream
DataStream
What are the thread-safety guarantees of the public API?
(explicitly define whether it is thread-safe or not)

If applicable, clarifying these points would make the design much easier to understand and help avoid misuse.

xianshijing-lk:
Do you find it useful to have the agents.md ? 

like, if claude / codex / cursor generate better results with this agents.md ? or it does not have much impact?

AGENTS.md
alan-george-lk: We as a team should come up with a `.clang-format` style we all like, then the `cpp-linter` action I am using in the `clang-tidy` PR can actually run/enforce that for us. But as far as this file, we could point the Agent to the `.clang-format` file once that becomes a thing. Probably a separate scope of work beyond `clang-tidy`, easy to implement but harder to align on preferences ;) 

See here: https://clang.llvm.org/docs/ClangFormat.html

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+## Project Overview
+
+This is **client-sdk-cpp**, the official LiveKit C++ client SDK. It wraps a Rust core (`client-sdk-rust/`) via a protobuf-based FFI bridge. All WebRTC, networking, and media logic lives in Rust; the C++ layer provides an ergonomic API for C++ consumers.
+
+## Architecture
+
+### Core Principle
+Rust owns as much of the business logic as possible. If a feature may be used by another SDK it should be implemented in Rust. Since this is an SDK,
+ensure backwards compatibility is maintained when possible.
```
alan-george-lk, line 11: Maybe explicitly list to avoid breaking changes in headers, or if absolutely necessary, prompt the user/flag them explicitly and require opt-in before proceeding?

Edit: looks like this is called out later in the file

AGENTS.md
alan-george-lk: I don't know if the Agent would care, but it's good MD practice generally to just do:

```
1.
1.
1.
```

and then it'll render the right order

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+SDK features follow pattern:
+
+1. Serialize a protobuf `FfiRequest` message.
+2. Send it to Rust via `FfiClient::instance().sendRequest(req)`.
+3. Receive a synchronous `FfiResponse` or an asynchronous `FfiEvent` callback.
+4. Expose the result through the C++ API.
+
+When making larger scale changes, check with the developer before committing to architecture changes involving changes to the `client-sdk-rust/` submodule.
+
+### Directory Layout
```
alan-george-lk, line 68: Is this a required feature of this file? I fear this will get out of date quickly.

Or maybe we add a clause saying "whenever directory structure changes, update AGENTS.md file"

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+
+- **`FfiClient`** — Singleton that sends FFI requests to Rust and dispatches event callbacks. Defined in `src/ffi_client.h` (internal).
+- **`FfiHandle`** — RAII wrapper for Rust-owned resource handles; drop releases the FFI resource.
+- **`Room`** — Central object for connecting to a LiveKit server room.
+- **`RoomDelegate`** — Virtual callback interface for room lifecycle events.
+- **`SubscriptionThreadDispatcher`** — Owns callback registrations and per-subscription reader threads for audio, video, and data tracks. `Room` delegates callback management here.
+- **`LocalParticipant` / `RemoteParticipant`** — Participant objects for publishing and receiving tracks.
+
+## Build
+for building, use the build.sh script for Linux and macOS, and the build.cmd script for Windows. Do not invoke CMake directly.
```
alan-george-lk: Nit: fully valid markdown has new lines between section headers/content. There's a nice `markdownlint` VSCode plugin that I use to help with this stuff: https://marketplace.cursorapi.com/items/?itemName=DavidAnson.vscode-markdownlint

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+
+```
+./build.sh debug              # Debug build
```
alan-george-lk, line 98: Similar comment to the repo directory structure, might be good to call out any adjustments to these script args should be updated in READMEs and this agents file

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Logging
+
+- Use `LK_LOG_*` macros from `src/lk_log.h` (internal header, **not** public API).
+- `LK_LOG_ERROR` — when something is broken.
+- `LK_LOG_WARN` — when something is unexpected but recoverable.
+- `LK_LOG_INFO` — initialization, critical state changes.
+- `LK_LOG_DEBUG` — potentially useful diagnostic info.
+- Do not spam logs. Keep it concise.
+- In production/internal SDK code, do not use `std::cout`, `std::cerr`, `printf`, or raw spdlog calls; use the logging APIs instead. Tests may use standard streams sparingly for ad hoc diagnostics when appropriate.
```
alan-george-lk, line 133: Good call out -- in my experience I also like using logging APIs for unit tests, that way you can add more detail at a debug/trace level and keep that off for production/CI runs.

AGENTS.md
```diff
@@ -0,0 +1,158 @@
...
+### Public API Surface
+
+- Keep public APIs small. Minimize what goes into `include/livekit/`.
+- Never introduce backwards compatibility breaking changes to the public API.
+- `lk_log.h` lives under `src/` (internal). The public-facing logging API is `include/livekit/logging.h`.
+- spdlog must not appear in any public header or installed header.
+
+### Error Handling
+
+- Throw exceptions for broken/unrecoverable states.
```
alan-george-lk: This is probably a larger team discussion: I am fairly anti-exception for robotics/embedded applications. Some safety certification specifications explicitly disallow exceptions. This would apply more to AV/aerospace customers of LiveKit due to safety requirements.

It would be unfortunate to lose a customer opportunity due to C++ feature usage like this. Looking at the existing SDK code we may not be able to undo existing exceptions, but for new code I'd prefer not to use them when possible.

`bool`, `std::optional`, `Expected<Type, Error>` ([in STL later](https://en.cppreference.com/w/cpp/utility/expected.html)), etc. (which I think we're already using in places)

AGENTS.md
stephen-derosa: ok i will remove this comment for now. If this gets in before the clang format PR can you be sure to update this AGENTS.md file?

AGENTS.md
stephen-derosa: i ended up just using bullets to try to not introduce confusion to the agents.

AGENTS.md
alan-george-lk: Yup sure can

AGENTS.md
stephen-derosa: good point, added a not to update on changes.

AGENTS.md
stephen-derosa: done!

AGENTS.md
stephen-derosa: Im curious to know more of the "Some safet certification specifications explicitily disallow exceptions". I could think of a car rolling on the street, exception, car doesnt stop. ha.

In general i agree with the sentiment of dont use exceptions if you dont have to, this line is definitely bad. 

I could change this to "Do not throw exceptions which must be caught by the application. Instead, prefer bool, std::optional, or a custom implementation of Result<Type, Error>".

Curious on your thoughts as well @xianshijing-lk

AGENTS.md
stephen-derosa: sweet thanks! (really wish the developers name was mark or marcella lol)

AGENTS.md
stephen-derosa: good call out -- updated!

stephen-derosa: updated, lmk what you think

AGENTS.md
stephen-derosa: I have the `core principals` section:
```
### Core Principle
Rust owns as much of the business logic as possible. If a feature may be used by another SDK it should be implemented in Rust. Since this is an SDK, ensure backwards compatibility is maintained when possible.
```

I added "The C++ layer should be a thin wrapper around the Rust core. " after the first sentence.

AGENTS.md
alan-george-lk: The idea is you avoid using them as much as possible in your own code (some third-party libraries like STL, ROS, etc. may throw), and handle errors using easy to follow control flow (overhead cost incurred by that, tradeoffs).

AGENTS.md
stephen-derosa: WRT threading, added [this in this commit](https://github.com/livekit/client-sdk-cpp/pull/91/changes/e434d343b1b5e04401f98fd2d14098e1fe9364ec). In general i was trying not to provide too much information to avoid confusion, but i think this is a good call out.

stephen-derosa:
> Do you find it useful to have the agents.md ?
> 
> like, if claude / codex / cursor generate better results with this agents.md ? or it does not have much impact?

@xianshijing-lk yeah i have found this file to be useful. Its hard to quantify but i have noticed the rules taking place such as copyright and logging preferences.

xianshijing-lk (approved):
lgtm, thanks Stephen

PR #92, opened 2026-04-10
docker base image for faster builds

stephen-derosa:
Breakout Dockerfile into .base and .sdk docker files. They build the base and sdk respectively.

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -1,0 +1,363 @@
+name: Docker Images
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - src/**
+      - include/**
+      - bridge/**
+      - client-sdk-rust/**
+      - CMakeLists.txt
+      - build.sh
+      - build.cmd
+      - build.h.in
+      - CMakePresets.json
+      - cmake/**
+      - data/**
+      - docker/Dockerfile.base
+      - docker/Dockerfile.sdk
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  detect-changes:
+    name: Detect Docker image changes
+    runs-on: ubuntu-latest
+    outputs:
+      base_changed: ${{ steps.changes.outputs.base_changed }}
+      sdk_changed: ${{ steps.changes.outputs.sdk_changed }}
+      base_hash: ${{ steps.hash.outputs.base_hash }}
+      base_image: ${{ steps.refs.outputs.base_image }}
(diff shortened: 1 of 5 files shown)


.github/workflows/docker-images.yml
```diff
@@ -0,0 +1,363 @@
...
+          base_hash="$(shasum -a 256 docker/Dockerfile.base | awk '{print substr($1,1,12)}')"
+          echo "base_hash=${base_hash}" >> "$GITHUB_OUTPUT"
+
+      - name: Detect changed inputs
+        id: changes
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          if [[ "${{ github.event.before }}" == "0000000000000000000000000000000000000000" ]]; then
```
alan-george-lk, line 64: Great idea overall, my main feedback is there are a lot of new stages and scripting to achieve the result; I am spinning up on the nuances of GitHub actions and refreshing my understanding of Docker, but I am wondering if there's a simpler caching detection mechanism within Docker itself. 

Will other SDKs need a similar flow for their docker image support?

alan-george-lk (approved):

.github/workflows/docker-images.yml
```diff
@@ -0,0 +1,363 @@
...
+      - client-sdk-rust/**
+      - CMakeLists.txt
+      - build.sh
+      - build.cmd
+      - build.h.in
+      - CMakePresets.json
+      - cmake/**
+      - data/**
+      - docker/Dockerfile.base
+      - docker/Dockerfile.sdk
```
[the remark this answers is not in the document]
stephen-derosa: no scheduled build for now to avoid bloat. If we see a need for it, its easy enough to add.

.github/workflows/docker-images.yml
stephen-derosa: by default docker caches its build process and layers, but since the runners are publicly hosted we have no guarantee to their lifecycle/caching so we dont really get to utilize it. I do agree the scripting is a little more than i would like to split it out, but the work is relatively concise to: resolving image name/tag, ensure there was a diff on the file to trigger a build, building all architectures, then verifying the installs. If theres a better way to do it im totally game :)


WRT other SDKs: if they want to dockerize they will probably have a similar workflow. It gets hairy trying to build out a single base image used across multiple SDKs without bloating the base image. That being said im sure theres some greatest common denominator. Luckily for them other languages have a much easier time with deps than cpp so Docker hasnt been a push for them 😅

PR #94, opened 2026-04-13
enable single peer connection as default

xianshijing-lk:


include/livekit/room.h
```diff
@@ -79,7 +79,7 @@ struct RoomOptions {
   // Enable single peer connection mode. When true, uses one RTCPeerConnection
   // for both publishing and subscribing instead of two separate connections.
   // Falls back to dual peer connection if the server doesn't support single PC.
-  bool single_peer_connection = false;
+  bool single_peer_connection = true;
```
alan-george-lk, line 82: Is this technically a breaking change?

alan-george-lk (approved):

include/livekit/room.h
xianshijing-lk: No, it is just using different implementation for signaling.

Our developers should not notice any difference except for a smaller connection latency

PR #97, opened 2026-04-15
Add clang-tidy support

alan-george-lk:
This PR adds support for `clang-tidy` static analysis within the CI. The list of initial checks can be found in the `.clang-tidy` file at the root of the repository.

Fixes to `.h/.cpp` files flagged by the analysis were also done.


src/room_proto_converter.cpp
```diff
@@ -57,7 +57,6 @@ DataPacketKind toDataPacketKind(proto::DataPacketKind in) {
   case proto::KIND_LOSSY:
     return DataPacketKind::Lossy;
   case proto::KIND_RELIABLE:
-    return DataPacketKind::Reliable;
```
xianshijing-lk, line 60: can we remove the default instead ? that will allow us to catch the problem in compilation

include/livekit/audio_stream.h
```diff
@@ -142,7 +142,7 @@ class AudioStream {
   FfiHandle stream_handle_;
 
   // Listener id registered on FfiClient
-  std::int64_t listener_id_{0};
+  std::int32_t listener_id_{0};
```
xianshijing-lk, line 145: Is this change intentional ? 
Didn't you mention that you would like to change the listener id to int64_t ?

.github/workflows/builds.yml
```diff
@@ -545,3 +546,86 @@ jobs:
...
+      pull-requests: write
+
+    steps:
+      - name: Checkout (with submodules)
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: recursive
+          fetch-depth: 1
+
+      - name: Install dependencies
```
stephen-derosa, line 565: holy deps of clang, im surprised there isnt a clanh-tidy ez unstall

bridge/include/livekit_bridge/rpc_constants.h
stephen-derosa, line 1: ugh sorry you have to drag in changes to the bridge

include/livekit/audio_stream.h
stephen-derosa, line 145: why are we reducing size here?

include/livekit/data_stream.h
```diff
@@ -78,7 +78,7 @@ struct ByteStreamInfo : BaseStreamInfo {
 class TextStreamReader {
 public:
   /// Construct a reader from initial stream metadata.
-  explicit TextStreamReader(const TextStreamInfo &info);
+  explicit TextStreamReader(TextStreamInfo info);
```
stephen-derosa, line 81: <3

build.sh
```diff
@@ -408,6 +410,32 @@ case "${cmd}" in
...
+    configure
+    build
+    if [[ "${DO_BUNDLE}" == "1" ]]; then
+      install_bundle
+      if [[ "${DO_ARCHIVE}" == "1" ]]; then
+        archive_bundle
+      fi
+    fi
+    ;;
+  debug-all)
```
xianshijing-lk, line 426: are these changes belonging to this clang-tidy PR ? 

Or you need to rebase ?

include/livekit/audio_stream.h
stephen-derosa: ffi returns a int32_t?

include/livekit/subscription_thread_dispatcher.h
```diff
@@ -404,7 +404,7 @@ class SubscriptionThreadDispatcher {
   /// is extracted and returned to the caller for joining outside the lock.
   std::thread startAudioReaderLocked(const CallbackKey &key,
                                      const std::shared_ptr<Track> &track,
-                                     AudioFrameCallback cb,
+                                     const AudioFrameCallback &cb,
```
stephen-derosa, line 407: hm, not sure why these werent const before, thanks!

src/tests/stress/test_latency_measurement.cpp
```diff
@@ -466,7 +465,8 @@ TEST_F(LatencyMeasurementTest, AudioLatency) {
         uint64_t send_time_us = last_high_energy_send_time_us.load();
 
         if (send_time_us > 0) {
-          double latency_ms = (receive_time_us - send_time_us) / 1000.0;
+          double latency_ms =
```
stephen-derosa, line 468: can this be const? And if so, shouldnt clang be finding that?

src/tests/stress/test_latency_measurement.cpp
```diff
@@ -692,7 +692,8 @@ TEST_F(LatencyMeasurementTest, FullDeplexAudioLatency) {
       if (send_from_a_us == 0 || detect_us <= send_from_a_us) {
         continue;
       }
-      double a_to_b_ms = (detect_us - send_from_a_us) / 1000.0;
+      double a_to_b_ms =
```
stephen-derosa, line 695: same const comment

src/tests/stress/test_latency_measurement.cpp
```diff
@@ -766,7 +767,8 @@ TEST_F(LatencyMeasurementTest, FullDeplexAudioLatency) {
       }
       a_received_pulse_id.store(pulse_id);
 
-      double b_to_a_ms = (receive_us - send_from_b_us) / 1000.0;
+      double b_to_a_ms =
```
stephen-derosa, line 770: same const comment

src/tests/stress/test_latency_measurement.cpp
```diff
@@ -777,7 +779,8 @@ TEST_F(LatencyMeasurementTest, FullDeplexAudioLatency) {
       }
 
       if (send_from_a_us > 0) {
-        double rtt_ms = (receive_us - send_from_a_us) / 1000.0;
+        double rtt_ms =
```
stephen-derosa, line 782: same const comment

src/audio_frame.cpp
```diff
@@ -54,8 +54,7 @@ AudioFrame AudioFrame::create(int sample_rate, int num_channels,
   const std::size_t count = static_cast<std::size_t>(num_channels) *
                             static_cast<std::size_t>(samples_per_channel);
   std::vector<std::int16_t> data(count, 0);
-  return AudioFrame(std::move(data), sample_rate, num_channels,
-                    samples_per_channel);
+  return {std::move(data), sample_rate, num_channels, samples_per_channel};
```
stephen-derosa: is this enforced by clang? I personally prefer an explicit type here. nbd either way.

src/audio_stream.cpp
```diff
@@ -212,8 +212,7 @@ void AudioStream::onFfiEvent(const FfiEvent &event) {
   }
   if (ase.has_frame_received()) {
     const auto &fr = ase.frame_received();
-    AudioFrame frame = AudioFrame::fromOwnedInfo(fr.frame());
-    AudioFrameEvent ev{std::move(frame)};
+    AudioFrameEvent ev{AudioFrame::fromOwnedInfo(fr.frame())};
```
stephen-derosa, line 215: personal preference or clang?

src/data_stream.cpp
```diff
@@ -189,21 +191,22 @@ bool ByteStreamReader::readNext(std::vector<std::uint8_t> &out) {
 // =====================================================================
 
 BaseStreamWriter::BaseStreamWriter(
-    LocalParticipant &local_participant, const std::string &topic,
-    const std::map<std::string, std::string> &attributes,
-    const std::string &stream_id, std::optional<std::size_t> total_size,
-    const std::string &mime_type,
-    const std::vector<std::string> &destination_identities,
-    const std::string &sender_identity)
+    LocalParticipant &local_participant, std::string topic,
```
stephen-derosa, line 194: id be interested to understand why clang prefers this move semantic rather than passing in a const ref

src/data_track_stream.cpp
```diff
@@ -58,7 +58,7 @@ bool DataTrackStream::read(DataTrackFrame &out) {
   std::unique_lock<std::mutex> lock(mutex_);
   cv_.wait(lock, [this] { return frame_.has_value() || eof_ || closed_; });
 
-  if (closed_ || (!frame_.has_value() && eof_)) {
+  if (closed_ || !frame_.has_value()) {
```
stephen-derosa: why are we no longer checking for eof_?

src/ffi_client.cpp
```diff
@@ -244,7 +247,7 @@ void FfiClient::PushEvent(const proto::FfiEvent &event) const {
 
 void LivekitFfiCallback(const uint8_t *buf, size_t len) {
   proto::FfiEvent event;
-  event.ParseFromArray(buf, len);
+  event.ParseFromArray(buf, static_cast<int>(len)); // TODO: this fixes for now, what if len exceeds int?
```
stephen-derosa, line 250: hm if len > max int than the ffi callback would be receiving a buffer size of ~2 gb, which i think would be a real problem, which theoretically should never happen idt.

it might worth putting a check in if `len > std::numeric_limits<int>::max()` then error

src/lk_log.h
```diff
@@ -20,8 +20,7 @@
 #include <memory>
 #include <spdlog/spdlog.h>
 
-namespace livekit {
-namespace detail {
+namespace livekit::detail {
```
stephen-derosa, line 23: at one point i liked the `detail` namespace...now i think its stupid. Definetly no need to add awork but if you agree we can consider changing

src/room.cpp
```diff
@@ -191,7 +191,7 @@ bool Room::Connect(const std::string &url, const std::string &token,
 
     // Install listener (Room is fully initialized)
     auto listenerId = FfiClient::instance().AddListener(
-        std::bind(&Room::OnEvent, this, std::placeholders::_1));
+        [this](const proto::FfiEvent &e) { OnEvent(e); });
```
stephen-derosa, line 194: clang prefers a lambda over std::bind?

src/room.cpp
```diff
@@ -300,7 +300,7 @@ void Room::setOnVideoFrameCallback(const std::string &participant_identity,
                                    VideoStream::Options opts) {
   if (subscription_thread_dispatcher_) {
     subscription_thread_dispatcher_->setOnVideoFrameCallback(
-        participant_identity, source, std::move(callback), std::move(opts));
+        participant_identity, source, std::move(callback), opts);
```
stephen-derosa, line 303: how come we no longer need to std::move(opts)?

src/room_proto_converter.cpp
```diff
@@ -57,7 +57,6 @@ DataPacketKind toDataPacketKind(proto::DataPacketKind in) {
   case proto::KIND_LOSSY:
     return DataPacketKind::Lossy;
   case proto::KIND_RELIABLE:
```
stephen-derosa: why dont we need this return anymore?

src/stats.cpp
```diff
@@ -50,7 +50,6 @@ QualityLimitationReason fromProto(livekit::proto::QualityLimitationReason r) {
   case P::LIMITATION_BANDWIDTH:
     return QualityLimitationReason::Bandwidth;
   case P::LIMITATION_OTHER:
```
stephen-derosa: why dont we need the return anymore?

src/room_proto_converter.cpp
stephen-derosa: I see its also the return of the default case, but doesnt the default only get hit if there was no valid `case` ?

src/subscription_thread_dispatcher.cpp
```diff
@@ -212,11 +212,10 @@ void SubscriptionThreadDispatcher::handleTrackSubscribed(
   std::thread old_thread;
   {
     std::lock_guard<std::mutex> lock(lock_);
-    if (track->kind() == TrackKind::KIND_AUDIO &&
-        audio_callbacks_.find(key) == audio_callbacks_.end()) {
-      key = fallback_key;
-    } else if (track->kind() == TrackKind::KIND_VIDEO &&
-               video_callbacks_.find(key) == video_callbacks_.end()) {
+    if ((track->kind() == TrackKind::KIND_AUDIO &&
```
stephen-derosa, line 215: FMU is this personal preference or clang?

src/subscription_thread_dispatcher.cpp
```diff
@@ -498,15 +497,15 @@ std::thread SubscriptionThreadDispatcher::startAudioReaderLocked(
 
   ActiveReader reader;
   reader.audio_stream = stream;
-  auto stream_copy = stream;
   const std::string participant_identity = key.participant_identity;
   const TrackSource source = key.source;
+  // NOLINTBEGIN(bugprone-lambda-function-name)
   reader.thread =
-      std::thread([stream_copy, cb, participant_identity, source]() {
+      std::thread([stream, cb, participant_identity, source]() {
```
stephen-derosa, line 504: if we dont need stream_copy, can we make `stream` const?

src/video_frame.cpp
```diff
@@ -46,29 +46,25 @@ std::size_t computeBufferSize(int width, int height, VideoBufferType type) {
     return w * h * 4;
 
   case VideoBufferType::RGB24:
```
stephen-derosa, line 48: is this clang enforced? for simple returns of an enum the removing of a return is somewhat readable, but i find this rather confusing

src/video_frame.cpp
```diff
@@ -46,29 +46,25 @@ std::size_t computeBufferSize(int width, int height, VideoBufferType type) {
...
-
   case VideoBufferType::I444:
-    // Y, U, V all full resolution
+    // 3 bytes per pixel (RGB24: packed; I444: Y+U+V all full resolution)
     return w * h * 3;
 
   case VideoBufferType::I420:
-  case VideoBufferType::NV12:
+    // Y (1 byte) + U (1 byte) + V (1 byte)
+  case VideoBufferType::NV12: {
```
stephen-derosa, line 55: why are {} needed for this case?

build.sh
```diff
@@ -34,9 +34,11 @@ Commands:
   debug             Configure + build Debug version (build-debug/)
   debug-examples    Configure + build Debug version with examples
   debug-tests       Configure + build Debug version with tests
+  debug-all         Configure + build Debug version with tests + examples
   release           Configure + build Release version (build-release/)
   release-examples  Configure + build Release version with examples
   release-tests     Configure + build Release version with tests
+  release-all       Configure + build Release version with tests + examples
```
stephen-derosa, line 41: nice!

stephen-derosa (commented):
in general this looks good, thanks for doing! I have some questions/comments, nothing major

include/livekit/audio_stream.h
alan-george-lk: This was intentional: our options are to move this to `int64_t`: https://github.com/livekit/client-sdk-cpp/blob/e69c646d355ce41de59c2aade09a70863699d3d2/src/ffi_client.h#L68

Or adjust the other listener ID types internally.

I opted for the latter option, because `ffi_client.h` and this typedef are public API, the other instances in the streams are private member variables. Open to discussion

build.sh
alan-george-lk: Intentional: the `release-all` is now used to build the code and tests for `clang-tidy` parsing, the `debug-all` is just to have feature parity on our build options

.github/workflows/builds.yml
alan-george-lk: I might be able to clean this up -- standby

bridge/include/livekit_bridge/rpc_constants.h
alan-george-lk: There's a valid argument to ignore the bridge paths for clang-tidy, but since the changes are already done... 🤷

include/livekit/audio_stream.h
alan-george-lk: See this comment: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3102262996

src/subscription_thread_dispatcher.cpp
alan-george-lk: This was a clang flag -- it flags duplicated code with different conditions, same thing for switch statements (also changed in this PR). Simple example:

```c++
if (cond_a)
{
    doThing();
else if (cond_b)
{
    doThing();
}
```

Clang flags and suggests:

```c++
if (cond_a || cond_b)
{
    doThing();
}
```

src/data_track_stream.cpp
alan-george-lk: I was also confused at first by this one, but it makes sense:

- The conditional variable has three exit conditions: `frame_.has_value()`, `eof_`, `closed_`
- On line `61`, if `closed_` is false and `frame.has_value()` is false, the only reason we got to this point was because `eof_` was `true` so it's a redundant `&&` clause

This is an incredibly minor optimization, it's an easy candidate for a `// NOLINT` if you're worried about changing behavior, I was concerned about this one too

src/tests/stress/test_latency_measurement.cpp
alan-george-lk: Fantastic catch: `misc-const-correctness` is not on currently, and is actually one of my favorite ones since it helps enforce `const` usage which can prevent unintended modification of variables. 

This one could be noisy -- how do we feel about turning it on in a separate PR since presumably there will be many changes alongside it?

src/audio_frame.cpp
alan-george-lk: Yup, we can undo if you prefer: https://clang.llvm.org/extra/clang-tidy/checks/modernize/return-braced-init-list.html

src/tests/stress/test_latency_measurement.cpp
alan-george-lk: See here: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3102400605

src/tests/stress/test_latency_measurement.cpp
alan-george-lk: See here: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3102400605

src/tests/stress/test_latency_measurement.cpp
alan-george-lk: See here: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3102400605

src/audio_stream.cpp
alan-george-lk: Clang: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3102405463

src/data_stream.cpp
alan-george-lk: See here: https://clang.llvm.org/extra/clang-tidy/checks/modernize/pass-by-value.html

src/video_frame.cpp
alan-george-lk: Whenever a `case` implementation is defining variables or has more than a simple statement, it's good practice to scope it with `{}`. In this case, the variables `chroma_w` and `chroma_h` are used across multiple cases, so this scopes them to avoid variable duplication compilation errors

src/video_frame.cpp
alan-george-lk: As mentioned in another comment, clang is going to nitpick us if we have more than one instance of the same code executing with different conditions. This is very debatable and am happy to go back on it. See here for a similar example: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3102279624

Basically these switch cases had different conditions but the same code, clang is suggesting to collapse them into cascading conditions (if either of these cases happen, do this)

src/subscription_thread_dispatcher.cpp
alan-george-lk: Made `stream` `const` locally, will push all changes up together

src/lk_log.h
alan-george-lk: I've seen `detail` used previously, I think it has some value for headers if you truly want to hide implementation for code auto-complete for example, I could go either way

src/room.cpp
alan-george-lk: Yup: https://clang.llvm.org/extra/clang-tidy/checks/modernize/avoid-bind.html

src/room.cpp
alan-george-lk: clang-tidy flagged this because `opts` is trivially-copyable (no heap allocations in its members), so `std::move` isn't really optimizing anything and could be misleading that a proper move is happening

src/room_proto_converter.cpp
alan-george-lk: This is another example of this: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3102279624

Since @xianshijing-lk also commented on this, I am wondering if we disable this check. It's not adding much value.

src/stats.cpp
alan-george-lk: See my reply here: https://github.com/livekit/client-sdk-cpp/pull/97#discussion_r3101952020

src/ffi_client.cpp
alan-george-lk: Agreed, would that be an exception throw in this case (shudders)?

I think the correct solution is to modify this to be a `int len` instead, but that could break things technically downstream: https://github.com/livekit/client-sdk-cpp/blob/main/src/ffi_client.h#L62

Thoughts @xianshijing-lk ?

src/tests/stress/test_latency_measurement.cpp
stephen-derosa: discussed offline, lets make const correctness the "next" MR

src/audio_frame.cpp
stephen-derosa: that would be my preference, not strong thoughts -- after all the return is typed. Your choice :)

src/data_track_stream.cpp
stephen-derosa: hm, then why do we need to check `closed_` and `!frame.has_value()` again if they are in the cv.wait() ?

src/lk_log.h
stephen-derosa: if its not unique, lets just leave it :)

src/room_proto_converter.cpp
stephen-derosa: my vote would be to disable this check

alan-george-lk: Upon further discussion, deciding to keep this check in

src/data_track_stream.cpp
alan-george-lk: In the case that `eof_` fires `true` but the others aren't valid to proceed, because of all the `||` (or) clauses in that wait condition (I think, as a newbie)

src/room_proto_converter.cpp
alan-george-lk: I'm not following this comment -- could you elaborate?

This change comes from this check: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/branch-clone.html

The issue is flagged when two conditions lead to the same result.

.github/workflows/builds.yml
alan-george-lk: Reduced to critical ones only

src/audio_frame.cpp
alan-george-lk: Undone!

.github/workflows/builds.yml
```diff
@@ -545,3 +546,77 @@ jobs:
             cmake -S . -B build -DLIVEKIT_LOCAL_SDK_DIR=/opt/livekit-sdk
             cmake --build build --parallel
           '
+
+  clang-tidy:
```
xianshijing-lk, line 550: do we know how long this job will take ? 

It seems running the job on the whole code baes? I wonder if we should gate it with 
Set files-changed-only: true
or even consider lines-changed-only: true

src/room_proto_converter.cpp
xianshijing-lk: I think it is slight better to do 
switch (in) {
  case proto::KIND_LOSSY:
    return DataPacketKind::Lossy;
  case proto::KIND_RELIABLE: 
    return DataPacketKind::Reliable; 
  }

Rather than the default.

In that case if we adding a new value to the enum, like proto::KIND_BEST_EFFORT
WIth
switch (in) {
  case proto::KIND_LOSSY:
    return DataPacketKind::Lossy;
  case proto::KIND_RELIABLE: 
    return DataPacketKind::Reliable; 
  }
it will fail to compile. 

with 
switch (in) {
  case proto::KIND_LOSSY:
    return DataPacketKind::Lossy;
  default: 
    return DataPacketKind::Reliable; 
  }

The compilation will still work, but it is wrong.

src/stats.cpp
xianshijing-lk: same here, if that is OK, we should remove the default instead.

xianshijing-lk (approved):
One comment about the job.

xianshijing-lk (approved):
lgtm assuming you will address the comments.

src/data_track_stream.cpp
stephen-derosa: its confusing that in other places like `video_stream.cpp` and `audio_stream.cpp` we check:
```
if (closed_ || (queue_.empty() && eof_))
...
```
as well. Prefer keeping this in for alignment and potential bug introduction

.clang-tidy
stephen-derosa, line 1: please update AGENTS.md to include a line about formatting. I think you should be able to point it to this file and tell it to format and write code to adhere to these checks.

stephen-derosa (approved):
in general this looks good, thanks for doing!

src/room_proto_converter.cpp
alan-george-lk: The tricky thing here is it's a warning not an error in C++, therefore until we do `-Werror` this could / will likely go through silently. I removed the default case locally and it still compiled. Which seems risky to introduce.

I'd like to treat warnings as errors in a future PR, but that's bigger undertaking. What if we `//NOLINT` these for now and come back to it?

include/livekit/audio_stream.h
xianshijing-lk: Thanks, keeping the public API unchanged is slightly preferred as well.

src/data_track_stream.cpp
alan-george-lk: Undone

.github/workflows/builds.yml
alan-george-lk: Waiting until all the checks/PR state is happy -- then will flip to the diff-based approach before merging to main

.clang-tidy
alan-george-lk: Formatting is separate -- but can update file for writing code to these checks

alan-george-lk: Added

.github/workflows/builds.yml
alan-george-lk: Changed to lines-changed-only

PR #101, opened 2026-04-16
Add code coverage reporting

alan-george-lk:
This PR does the following:

- Adds code coverage reporting (unit tests only) step to tests workflow
- This step adds a [summary markdown to the job run](https://github.com/livekit/client-sdk-cpp/actions/runs/25705184831#summary-75473645694), but also a full HTML report showing branching coverage of unit tests (example screenshot below)

Eventually we could host this HTML for view online somewhere, but for now it's easy enough to use the report.

Screenshot 2026-05-11 at 8 26 55 PM


xianshijing-lk (approved):
Nice stuff!!

stephen-derosa (approved):
LGTM!

stephen-derosa (commented):
not that we should but is there any way to fail if coverage is less than some amount?

PR #103, opened 2026-04-17
add instruction to add pre-commit clang-format hook

xianshijing-lk:
git does not track any file under .git/ folder, so we need to add some instruction for developers to setup their env manually

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,9 +481,10 @@ In some cases, you may need to perform a full clean that deletes all build artif
 ```
 
 ### Clang format
-CPP SDK is using clang C++ format
+CPP SDK is using clang C++ format, and set up the pre-commit hook (run once):
 ```bash
 brew install clang-format
+printf '#!/bin/sh\necho "Running clang-format..."\nFILES=$(git diff --cached --name-only -- "*.cpp" "*.cc" "*.h" "*.hpp")\n[ -z "$FILES" ] && exit 0\nfor f in $FILES; do clang-format -i "$f"; git add "$f"; done\necho "clang-format completed."\n' > .git/hooks/pre-commit && chmod +x .git/hooks/pre-commit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -481,9 +481,10 @@ In some cases, you may need to perform a full clean that deletes all build artif
 ```
 
 ### Clang format
-CPP SDK is using clang C++ format
+CPP SDK is using clang C++ format, and set up the pre-commit hook (run once):
 ```bash
 brew install clang-format
+printf '#!/bin/sh\necho "Running clang-format..."\nFILES=$(git diff --cached --name-only -- "*.cpp" "*.cc" "*.h" "*.hpp")\n[ -z "$FILES" ] && exit 0\nfor f in $FILES; do clang-format -i "$f"; git add "$f"; done\necho "clang-format completed."\n' > .git/hooks/pre-commit && chmod +x .git/hooks/pre-commit
```
stephen-derosa, line 487: 👍🏼

stephen-derosa (approved):

README.md
alan-george-lk, line 487: As an alternative (or an addition), could we add this to the build scripts? I.e. check if the hook exists, then create it if not? That way we could have better enforcement

alan-george-lk: Looks like there are some more automatic/elegant solutions, example: https://github.com/polymathrobotics/livekit_ros2_bridge/blob/main/.pre-commit-config.yaml

stephen-derosa:
LGTM anything holding this up @xianshijing-lk ?

alan-george-lk:
> LGTM anything holding this up @xianshijing-lk ?

@stephen-derosa I'm fine with this PR as-is, but there are some more robust options we may want to explore, see my linked comment above

PR #107, opened 2026-04-22
Clang-tidy: additional checks

alan-george-lk:
Follow on PR from #97 to enable `misc-const-correctness` check and apply fixes where it makes sense to.

Note that on an entire repo run for this check, there were some areas of the code that required minor refactors to achieve `const` on the flagged lines. I opted not to touch these areas (not worth risk of changing), and they're not flagged because the `cpp-linter` is only set to check changed lines.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,7 @@
 Checks: >
   -*,
   bugprone-*,
+  misc-const-correctness,
   performance-*,
   modernize-*,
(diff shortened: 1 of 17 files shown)


xianshijing-lk (approved):

stephen-derosa (approved):
awesome thanks for doing this! FMU: with this MR do we expect clang-tidy to return no errors?

PR #111, opened 2026-04-24
Benchmarks: Data tracks throughput

stephen-derosa:
`benchmark/data_track_throughput`: throughput executables and post processing tooling for visualizing Data Track MBPS throughput:

Sample plot:
image


xianshijing-lk:
Sorry, I didn't have time to go thru the details, but assuming this is done quickly and providing a way to benchmark / test our data track throughput. I like it. 

Assuming we don't need to over engineer to make such benchmark / test code perfect, I am giving a lgtm now.

xianshijing-lk (approved):

PR #116, opened 2026-05-05
clang-format in CI

alan-george-lk:
This PR adds the following:

- `scripts/clang-format.sh` that works similarly to `clang-tidy`, runs both locally and in CI with a report about files that need formatting
- Added `.clang-format` file based on `Google` style with some small adjustments to minimize overall change
- Ran formatting across all relevant files
- Updated documentation around how to run it locally/git hook instructions

## Testing

Tested thoroughly locally and in CI. Example output below.

When no issues are found:
```log
------------------------------------------------------------
clang-format summary: clean (0 of 109 file(s) needed formatting)
------------------------------------------------------------
```

Screenshot 2026-05-05 at 1 37 25 PM

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,2 @@
+---
+BasedOnStyle: LLVM
--- a/.clang-format
+++ b/.clang-format
@@ -1,2 +1,10 @@
 ---
-BasedOnStyle: LLVM
+BasedOnStyle: Google
+Language: Cpp
+Standard: c++17
+
+# LiveKit modifications to Google style below
+
+ColumnLimit: 120                  # more width on modern screens
+SpacesBeforeTrailingComments: 1   # one space for trailing namespace comments, e.g. `} // namespace foo` (Google uses 2)
+AccessModifierOffset: -2          # left-align public/protected/private with the class keyword (Google indents by 1)
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -17,6 +17,8 @@ on:
       - docker/Dockerfile.base
       - docker/Dockerfile.sdk
       - .github/workflows/**
+      - .clang-format
+      - scripts/clang-format.sh
   pull_request:
     branches: ["main"]
     paths:
@@ -33,6 +35,8 @@ on:
       - docker/Dockerfile.base
       - docker/Dockerfile.sdk
       - .github/workflows/**
+      - .clang-format
+      - scripts/clang-format.sh
   workflow_dispatch:
 
 permissions:
@@ -546,6 +550,51 @@ jobs:
             cmake -S . -B build -DLIVEKIT_LOCAL_SDK_DIR=/opt/livekit-sdk
             cmake --build build --parallel
           '
+
+  clang-format:
+    name: clang-format
+    runs-on: ubuntu-latest
+    continue-on-error: false
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          # No submodules: scripts/clang-format.sh only walks our own src/
+          # tree (the client-sdk-rust submodule has its own .clang-format).
+          fetch-depth: 1
+
+      - name: Install clang-format 19
+        run: |
+          set -eux
+          # Pin clang-format 19 to match the clang-tidy job and keep CI
+          # output deterministic across runs. Ubuntu 24.04's default
+          # clang-format ships with LLVM 18; pull 19 from the upstream
+          # LLVM apt repository so a developer running clang-format-19
+          # locally gets the same result CI does.
+          sudo install -m 0755 -d /etc/apt/keyrings
+          wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key \
+            | sudo tee /etc/apt/keyrings/llvm.asc >/dev/null
+          sudo chmod a+r /etc/apt/keyrings/llvm.asc
+          codename=$(lsb_release -cs)
+          echo "deb [signed-by=/etc/apt/keyrings/llvm.asc] http://apt.llvm.org/${codename}/ llvm-toolchain-${codename}-19 main" \
+            | sudo tee /etc/apt/sources.list.d/llvm-19.list >/dev/null
+          sudo apt-get update
+          sudo apt-get install -y clang-format-19
+          sudo ln -sf /usr/bin/clang-format-19 /usr/local/bin/clang-format
+          clang-format --version
+
+      - name: Run clang-format
+        env:
+          FORMAT_BLOB_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+        # This script is intended to be run locally and in CI. It will
+        # auto-detect the GHA environment and emit PR file annotations
+        # plus a markdown step summary when GITHUB_ACTIONS=true. Default
+        # mode is check-only (--dry-run --Werror), which exits non-zero
+        # on any divergence from the repo-root .clang-format.
+        run: ./scripts/clang-format.sh
 
   clang-tidy:
     name: clang-tidy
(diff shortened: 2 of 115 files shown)


scripts/clang-format.sh
stephen-derosa, line 1: do we need to support a windows version of this?

src/video_frame.cpp
```diff
@@ -336,14 +336,18 @@ VideoFrame VideoFrame::fromOwnedInfo(const proto::OwnedVideoBuffer &owned) {
     std::size_t offset = 0;
     for (const auto &comp : info.components()) {
       const auto sz = static_cast<std::size_t>(comp.size());
-      const auto *src_ptr = reinterpret_cast<const std::uint8_t *>(comp.data_ptr()); // NOLINT(performance-no-int-to-ptr)
+      const auto *src_ptr =
+          // NOLINTNEXTLINE(performance-no-int-to-ptr)
```
stephen-derosa, line 336: nolintnextline in between the var and the value is a first for me lol, is it ok to have it the line above?

stephen-derosa (commented):
just doc-ing to ensure its tracked, need to update:
- AGENTS.md: `Use clang formatting that aligns with the existing codebase.`
- README.md: `clang-format: (coming soon) code formatting and style consistency`

src/video_frame.cpp
alan-george-lk: This was a column-width induced change, agreed upon larger `120` column width to avoid/clean this up

scripts/clang-format.sh
alan-george-lk: Right now we caveat in the read me that clang tools are just on Mac/Linux, I think clang-format does work on Windows since it doesn't depend on `compile_commands.json`, and this check would be a blocker for any Windows developers. What I don't want to do is duplicate/maintain the `clang-format.sh` script for a niche use case (windows based contributions). I think user on Windows could just run `clang-format.exe <path to offending files>` and wouldn't need the script (which is more convenient for us).

How do you feel about not duplicating the script and just adding a blurb to the readme saying you're on your own to run it? CI will still flag the offending files

.github/workflows/builds.yml
```diff
@@ -513,6 +517,45 @@ jobs:
             cmake --build build --parallel
           '
 
+  clang-format:
```
stephen-derosa, line 520: should this be in the new tests.yml  ? or its own format.yml? No real preference but i like that build.yml more or less just builds

.github/workflows/builds.yml
```diff
@@ -513,6 +517,45 @@ jobs:
...
+            | sudo tee /etc/apt/sources.list.d/llvm-19.list >/dev/null
+          sudo apt-get update
+          sudo apt-get install -y clang-format-19
+          sudo ln -sf /usr/bin/clang-format-19 /usr/local/bin/clang-format
+          clang-format --version
+
+      - name: Run clang-format
+        env:
+          FORMAT_BLOB_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+        run: ./scripts/clang-format.sh
```
stephen-derosa, line 554: <3

.clang-format
```diff
@@ -0,0 +1,10 @@
+---
+BasedOnStyle: Google
+Language: Cpp
+Standard: c++17
+
+# LiveKit modifications to Google style below
+
+ColumnLimit: 120                  # more width on modern screens
+SpacesBeforeTrailingComments: 1   # one space for trailing namespace comments, e.g. `} // namespace foo` (Google uses 2)
```
stephen-derosa, line 9: any reason to not use google standard of 2 here? I have no real pref

scripts/clang-format.sh
stephen-derosa: fine by me!

stephen-derosa (approved):
one suggestion, looks good to me!

.github/workflows/builds.yml
```diff
@@ -513,6 +517,45 @@ jobs:
...
+          sudo install -m 0755 -d /etc/apt/keyrings
+          wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key \
+            | sudo tee /etc/apt/keyrings/llvm.asc >/dev/null
+          sudo chmod a+r /etc/apt/keyrings/llvm.asc
+          codename=$(lsb_release -cs)
+          echo "deb [signed-by=/etc/apt/keyrings/llvm.asc] http://apt.llvm.org/${codename}/ llvm-toolchain-${codename}-19 main" \
+            | sudo tee /etc/apt/sources.list.d/llvm-19.list >/dev/null
+          sudo apt-get update
+          sudo apt-get install -y clang-format-19
+          sudo ln -sf /usr/bin/clang-format-19 /usr/local/bin/clang-format
```
xianshijing-lk: nit, I think this could conflict if runner image changes later
how about
export PATH="/usr/lib/llvm-19/bin:$PATH"
or just call 
clang-format-19 ?

scripts/clang-format.sh
```diff
@@ -0,0 +1,495 @@
...
+# own conventions (it's frozen pending removal), and examples/ has no
+# tracked C/C++ files today.
+files=()
+if (( ${#explicit_files[@]} > 0 )); then
+  files=("${explicit_files[@]}")
+else
+  while IFS= read -r -d '' path; do
+    files+=("${path}")
+  done < <(git ls-files -z \
+    'src/*.c' 'src/*.cc' 'src/*.cpp' 'src/*.cxx' \
```
xianshijing-lk: I wonder if this file glob coverage is  as recursive as intended, maybe use things like 
'src/**/*.cpp' 'include/**/*.h'

or something like 
git ls-files -z -- 'src' 'include' 'benchmarks' | grep ...

xianshijing-lk (approved):
lgtm with some comments

scripts/clang-format.sh
alan-george-lk: Added to README

.github/workflows/builds.yml
alan-george-lk: Right now both `clang-tidy` and `clang-format` are part of the build, I think realistically they should be in a `quality.yml` file or similar. Do you want me to attempt that here, or is it okay to move forward with this here since `clang-tidy` is already in here and we can move out these stages (+ a few others I have in mind) later?

.clang-format
alan-george-lk: I was going for less changes to the code compared to what we have now, if you'd rather have less changes to the clang format file, happy to drop this and re-run it

.github/workflows/builds.yml
alan-george-lk: I'm explicitly installing `clang-format-19`, so unless apt starts installing it in a weird place, it should always go to `/usr/bin/clang-format-19`. Am I understanding your concern right or no?

alan-george-lk: Oh I think I see, make the alias more specific. 

The reason I went with this was because the locally installed command (on Mac) is `clang-format`, which is invoked by the script. Otherwise the user would have to symlink or alias their `clang-format` to `clang-format-19` to run the script locally to match this, which would be ugly. I'm open to other ideas

.github/workflows/builds.yml
alan-george-lk: We decided to leave in for now

.clang-format
alan-george-lk: Deciding to leave it in

scripts/clang-format.sh
alan-george-lk: This does go recursive, I cleaned it up a bit since this comment, but it does hit things like `src/tests/*`

.github/workflows/builds.yml
alan-george-lk: This comment made me realize mac was not on clang-format-19, so updated version to match mac (clang-format-22)

xianshijing-lk: It is fine to keep the way as is

PR #113, opened 2026-04-29
Integration tests in CI

alan-george-lk:
This PR does the following:

- Moves all testing into a new `tests.yml` CI stage
- New more specific/performance improved runners for Windows/Mac
- Adds the ability to run integration tests in CI via a locally-installed LiveKit server (Linux only)
- Adjusts unstable integration tests to improve stability
  - PublishesAndReceivesFramesEndToEnd -> Adjusted to more closely match the Rust SDK layout
  - Timestamp test -> Matches the E2E TypeScript test by re-using the same timestamp throughout
- A more explicit parallel jobs function for improving CI times (has helped slightly on Windows, for example)
- Consolidates 5 different prior env vars for integration/stress tests down to `LIVEKIT_TOKEN_A/B`
- Changes Rust cache action to match Rust SDK
- Add timeouts to testing stages to catch deadlock bugs
- Disables GMock to save time on build (not used)

Some caveats:

- Still runs unit tests on all prior platforms
- Mac runner performance is not passing on C++ integration tests through FFI. Internal ticket was created to further debug this

PR #119, opened 2026-05-06
BugFix: Bot-343, DataTracks return from subscribe() and read() from rust sdk

stephen-derosa:
aligns with changes here:
https://github.com/livekit/rust-sdks/pull/1057

Depends on:
- [x] https://github.com/livekit/rust-sdks/pull/1057, need to update submodule
- [x] integration tests [in ci for validation](https://github.com/livekit/client-sdk-cpp/pull/113)

### DIFF
--- a/src/data_track_stream.cpp
+++ b/src/data_track_stream.cpp
@@ -65,6 +65,11 @@ bool DataTrackStream::read(DataTrackFrame& out) {
   return true;
 }
 
+std::optional<SubscribeDataTrackError> DataTrackStream::terminalError() const {
+  const std::scoped_lock<std::mutex> lock(mutex_);
+  return terminal_error_;
+}
+
 void DataTrackStream::close() {
   std::int32_t listener_id = -1;
   {
@@ -73,9 +78,14 @@ void DataTrackStream::close() {
       return;
     }
     closed_ = true;
+    // Preserve errors reported by EOS for post-stream inspection, but do not
+    // treat a local early close as a terminal subscription error.
+    if (!eof_) {
+      terminal_error_.reset();
+    }
     subscription_handle_.reset();
     listener_id = listener_id_;
-    listener_id_ = 0;
+    listener_id_ = -1;
   }
 
   if (listener_id != -1) {
@@ -103,7 +113,12 @@ void DataTrackStream::onFfiEvent(const FfiEvent& event) {
     DataTrackFrame frame = DataTrackFrame::fromOwnedInfo(fr);
     pushFrame(std::move(frame));
   } else if (dts.has_eos()) {
-    pushEos();
+    std::optional<SubscribeDataTrackError> error;
+    const auto& eos = dts.eos();
+    if (eos.has_error()) {
+      error = SubscribeDataTrackError::fromProto(eos.error());
+    }
+    pushEos(std::move(error));
   }
 }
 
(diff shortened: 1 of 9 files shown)


alan-george-lk (approved):
Looks great, love the bundled unit test

src/subscription_thread_dispatcher.cpp
```diff
@@ -724,6 +724,14 @@ std::thread SubscriptionThreadDispatcher::startDataReaderLocked(
         LK_LOG_ERROR("Data frame callback exception: {}", e.what());
       }
     }
+    const auto &error = stream->terminalError();
```
ladvoc: question: Is the user able to access the error reason when using the thread dispatcher?

ladvoc (approved):
LGTM, just the one question. Also, have you tested this against the Rust branch?

src/data_track_stream.cpp
```diff
@@ -75,6 +80,9 @@ void DataTrackStream::close() {
       return;
     }
     closed_ = true;
+    if (!eof_) {
+      terminal_error_.reset();
+    }
     subscription_handle_.reset();
     listener_id = listener_id_;
     listener_id_ = 0;
```
xianshijing-lk: just found out this might be a potential bug for double RemoveListener(listener_id);

when a close() is called, it will set listener_id_ = 0;
and you will do
if (listener_id != -1) {
    FfiClient::instance().RemoveListener(listener_id);
  }

That says, if close() is called twice, we will call FfiClient::instance().RemoveListener(0); which might be wrong

src/data_track_stream.cpp
```diff
@@ -67,6 +67,11 @@ bool DataTrackStream::read(DataTrackFrame &out) {
   return true;
 }
 
+std::optional<SubscribeDataTrackError> DataTrackStream::terminalError() const {
+  const std::scoped_lock<std::mutex> lock(mutex_);
+  return terminal_error_;
```
xianshijing-lk, line 70: [suggestion](https://pullpo.io/cc?l=suggestion) 

Is it intentional that terminal_error_ will still be return after close() ? or it should be valid only before calling close() ?

src/tests/unit/test_data_track_stream.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  DataTrackFrame frame;
+  EXPECT_FALSE(stream->read(frame));
+
+  const auto error = stream->terminalError();
+  ASSERT_TRUE(error.has_value());
+  EXPECT_EQ(error->code, SubscribeDataTrackErrorCode::UNPUBLISHED);
+  EXPECT_EQ(error->message, "track unpublished before subscription completed");
+}
+
```
xianshijing-lk, line 87: can you add more tests like 
relationship between Close before EOS ? like calling Close() before or after the EOS ? will terminal errors get preserved?
Duplicate EOS ? 
Events after close ? 
Receiving non data track events, what will happen ?

xianshijing-lk (approved):
lgtm if you address the comments

src/tests/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ if(UNIT_TEST_SOURCES)
     PRIVATE
       livekit
       spdlog::spdlog
-      $<$<PLATFORM_ID:Windows>:${LIVEKIT_PROTOBUF_TARGET}>
+      ${LIVEKIT_PROTOBUF_TARGET}
```
stephen-derosa, line 65: not sure why copilot thought this needed

src/data_track_stream.cpp
stephen-derosa: Yes, intentional. terminalError() is meant to be readable as post-stream state. If the stream has already reached EOS with an FFI-reported terminal error, close() preserves it. If close() is called locally before EOS, we clear terminal_error_ because the stream ended by local cancellation, not by a remote/FFI terminal failure.

I added a small comment here

src/subscription_thread_dispatcher.cpp
stephen-derosa: no in the current implementation

src/data_track_stream.cpp
stephen-derosa: good callout. listener_id_ should be -1. updated.

And we protect against superfluous close() calls with:
```
    if (closed_) {
      return;
    }
```

PR #121, opened 2026-05-07
Visibility controls on public API

alan-george-lk:
This PR changes the following:

- Define `LIVEKIT_API` macro to expose LiveKit API symbols across platforms (common convention, [example](https://github.com/abseil/abseil-cpp/blob/d0d4c599108be215585d7380d84a652ddd87607c/absl/base/config.h#L624-L653))
- All symbol visibility hidden by default otherwise
- Update relevant class/method symbols with this new macro
- Made include prefixes consistent for public API (all now use `#include "livekit/<blah"` convention, updated `AGENTS.md` to define this rule
- Hide third-party symbols from consumers
  - `spdlog` was installed at the system level in CI (and documented for users to install at that level), but was brought in via `FetchContent`/`vcpkg`. This is no longer needed (system install). Removed that flow and deprecated the CMake argument/functionality for using system level `spdlog`. It is now entirely within the SDK and fully private/no symbols exposed to clash

## Testing

- Built against/linked in downstream project that originally exposed this issue (suspected `spdlog` clash with third-party lib)
  - Without this fix and conflicting `spdlog`: `process has deied [pid 5915, exit code -7...`
- Added tester script to `tests.yml` CI stage to catch this across all OS moving forward, with the small caveat that we need to ensure any new deps are added to the symbol check
- Unit/integration tests are standalone binaries that practice this linking, ensured they still worked via CI

Screenshot 2026-05-11 at 11 23 44 AM


.github/workflows/builds.yml
```diff
@@ -133,15 +137,14 @@ jobs:
             libssl-dev \
             libprotobuf-dev protobuf-compiler \
             libabsl-dev \
-            libwayland-dev libdecor-0-dev \
-            libspdlog-dev
+            libwayland-dev libdecor-0-dev
```
stephen-derosa, line 140: im confused on why spdlog is no longer needed as a dep if we still need to compile against it

include/livekit/local_audio_track.h
```diff
@@ -22,6 +22,7 @@
 #include "audio_frame.h"
 #include "local_track_publication.h"
 #include "track.h"
+#include "visibility.h"
```
stephen-derosa: some places i see:
`#include "livekit/visibility.h"` in other locations `#include "visibility.h"`

.github/workflows/builds.yml
alan-george-lk: See PR body for a summary, but basically `FetchContent`/`vcpkg` in the CMake installs it, and we link it statically, so it's entirely contained to the build process. 

A system install like this (on linux for example) would put it somewhere like `/usr/lib/spdlog` and then during the linking process the compiler will use that version, which we don't want as a flow. I am pretty sure even before this change this install wasn't needed unless you configured the "use system spdlog" option, since again from the start of this comment we were already pulling it into the build

include/livekit/local_audio_track.h
alan-george-lk: Good call, I think Opus added it like this to match convention already in the files. We should come up with a standard convention (using the `livekit/` prefix or not) and commit to that. See how the ones above this line don't have the prefix, but this file does use them even though it's the same folder: https://github.com/livekit/client-sdk-cpp/blob/af0820c206cc3e3cb4356c72e792ec1ef4680778/include/livekit/track.h#L26

stephen-derosa: hm, this is probably AI run away, i think we should be using `#include "livekit/*`

alan-george-lk: I went with the `livekit/<blah>` prefix approach, it's more clear what the path is, doesn't depend on the build system exposing certain paths, and was the vast majority of cases compared to just `<blah>` include-path-wise.

I updated all the violating files and added to `AGENTS.md` what the rules are, lmk what you think.

.github/scripts/check_no_private_symbols.py
```diff
@@ -0,0 +1,247 @@
...
+    python3 check_no_private_symbols.py <path-to-shared-library>
+
+Optional environment variables:
+    LIVEKIT_SYMBOL_CHECK_VERBOSE=1   Print every leaked symbol (default: print
+                                    up to 20 examples).
+    LIVEKIT_SYMBOL_CHECK_EXTRA_FORBIDDEN=foo,bar   Additional comma-separated
+                                                   patterns to forbid.
+"""
+
+import os
```
alan-george-lk, line 40: This is an AI-generated script I prompted; I wanted to verify symbols weren't exported on other OSs/generally catch this issue again in the future

.github/workflows/builds.yml
```diff
@@ -181,6 +184,49 @@ jobs:
         shell: pwsh
         run: ${{ matrix.build_cmd }}
 
+      # ---------- Verify exported ABI: no spdlog/fmt/protobuf/absl leaks ----------
```
alan-george-lk: This can probably be removed, see: https://github.com/livekit/client-sdk-cpp/pull/121/changes#r3219803050

AGENTS.md
```diff
@@ -140,10 +140,55 @@ All source files must have the LiveKit Apache 2.0 copyright header. Use the curr
...
+  annotation because they emit no out-of-line symbols.
+- For internal symbols that must remain visible to in-tree tests (the tests
+  link the same shared library), use `LIVEKIT_INTERNAL_API`.  External
+  consumers must not rely on these.
+- Do not add `__declspec(dllexport)` or `__attribute__((visibility("default")))`
+  by hand; always go through `LIVEKIT_API` / `LIVEKIT_INTERNAL_API`.
+- On Windows, `WINDOWS_EXPORT_ALL_SYMBOLS` is **deliberately disabled** for the
+  `livekit` target.  Use `LIVEKIT_API` to export.
+
+CI enforces the policy via `scripts/check_no_private_symbols.py`, which is
```
alan-george-lk: Probably can be removed, see: https://github.com/livekit/client-sdk-cpp/pull/121/changes#r3219803050

.github/scripts/check_no_private_symbols.py
alan-george-lk: I moved this to `.github/scripts` since it's only invoked from CI in the tests workflow (can still be invoked locally). Better to have it than not have it

.github/workflows/builds.yml
```diff
@@ -9,6 +9,8 @@ on:
       - cpp-example-collection/**
       - bridge/**
       - client-sdk-rust/**
+      - cmake/**
```
alan-george-lk, line 12: This list is getting long, I wonder if we could/should have an exclude list instead, like `*.md`/other docs

stephen-derosa: definetly on board, up to you on including it this MR or not. looks like its as simple as:
```
    paths-ignore:
      - 'docs/**'
      - '**/*.md'
      - '.github/ISSUE_TEMPLATE/**'
      - '.gitignore'
      - ... 
```
`paths` -> `paths-ignore`

stephen-derosa (approved):
LGTM -- is there any test we can run to make sure we arent missing the `LIVEKIT_API` macro for items that should be public?

.github/workflows/tests.yml
```diff
@@ -27,6 +28,7 @@ on:
       - vcpkg.json
       - .token_helpers/**
       - .github/workflows/tests.yml
+      - .github/scripts/check_no_private_symbols.py
```
xianshijing-lk: do we want to run this check_no_private_symbols.py for every pull requests ? 

or should it be part of the release flow instead ? 

IMHO, one-time validation + occasional CI guard probably makes more sense for such check

alan-george-lk: Super easy to move it to the release CI flow, it's pretty lightweight though. Your call

.github/workflows/builds.yml
alan-george-lk: Will do on follow-on since this PR is getting large

.github/workflows/tests.yml
xianshijing-lk: Lets move it to release CI flow then

xianshijing-lk (approved):
lgtm assuming the check will be moved to release CI flow

.github/workflows/tests.yml
alan-george-lk: Moved

PR #122, opened 2026-05-09
Events Being Dropped on Connection

stephen-derosa:
# Overview
- Adds integration tests to validate events after connection
- `RoomReadyEventRequest` which gets called after ffi handle is created

# Testing
- new integration test
- manual addition of a sleep before the making the request shows that we still get the events

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ lib/
 web/
 *trace.json
 compile_commands.json
+user_stories
--- a/src/ffi_client.cpp
+++ b/src/ffi_client.cpp
@@ -35,7 +35,6 @@
 namespace livekit {
 
 namespace {
-
 inline void logAndThrow(const std::string& error_msg) {
   LK_LOG_ERROR("LiveKit SDK Error: {}", error_msg);
   throw std::runtime_error(error_msg);
--- a/src/room.cpp
+++ b/src/room.cpp
@@ -18,15 +18,11 @@
 
 #include <functional>
 
-#include "data_track.pb.h"
 #include "ffi.pb.h"
 #include "ffi_client.h"
 #include "livekit/audio_stream.h"
 #include "livekit/e2ee.h"
-#include "livekit/livekit.h"
-#include "livekit/local_data_track.h"
 #include "livekit/local_participant.h"
-#include "livekit/local_track_publication.h"
 #include "livekit/remote_audio_track.h"
 #include "livekit/remote_data_track.h"
 #include "livekit/remote_participant.h"
@@ -64,6 +60,16 @@ std::shared_ptr<livekit::RemoteParticipant> createRemoteParticipant(const proto:
   livekit::FfiHandle handle(static_cast<uintptr_t>(owned.handle().id()));
   return std::make_shared<livekit::RemoteParticipant>(std::move(handle), pinfo.sid(), pinfo.name(), pinfo.identity(),
                                                       pinfo.metadata(), std::move(attrs), kind, reason);
+}
+
+void readyForRoomEvent(std::uint64_t room_handle) {
+  FfiRequest req;
+  req.mutable_ready_for_room_event()->set_room_handle(room_handle);
+
+  const auto resp = FfiClient::instance().sendRequest(req);
+  if (!resp.has_ready_for_room_event()) {
+    throw std::runtime_error("FfiResponse missing ready_for_room_event");
+  }
 }
 
 } // namespace
@@ -118,12 +124,21 @@ bool Room::Connect(const std::string& url, const std::string& token, const RoomO
     }
     connection_state_ = ConnectionState::Reconnecting;
   }
-  auto fut = FfiClient::instance().connectAsync(url, token, options);
+
+  FfiClient::ListenerId listenerId = 0;
   try {
+    listenerId = FfiClient::instance().AddListener([this](const proto::FfiEvent& e) { OnEvent(e); });
+    {
+      const std::scoped_lock<std::mutex> g(lock_);
+      listener_id_ = listenerId;
+    }
+
+    auto fut = FfiClient::instance().connectAsync(url, token, options);
     auto connectCb = fut.get(); // fut will throw if it fails to connect to the room
 
     const auto& owned_room = connectCb.result().room();
     auto new_room_handle = std::make_shared<FfiHandle>(owned_room.handle().id());
+    const auto room_handle_id = static_cast<std::uint64_t>(new_room_handle->get());
     auto new_room_info = fromProto(owned_room.info());
 
     // Setup local particpant
@@ -147,6 +162,7 @@ bool Room::Connect(const std::string& url, const std::string& token, const RoomO
           std::make_unique<LocalParticipant>(std::move(participant_handle), pinfo.sid(), pinfo.name(), pinfo.identity(),
                                              pinfo.metadata(), std::move(attrs), kind, reason);
     }
+
     // Setup remote participants
     std::unordered_map<std::string, std::shared_ptr<RemoteParticipant>> new_remote_participants;
     {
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -34,3 +34,4 @@ lib/
 web/
 *trace.json
 compile_commands.json
+user_stories
```
alan-george-lk, line 37: What generates this path?

src/room.cpp
```diff
@@ -178,13 +180,28 @@ bool Room::Connect(const std::string& url, const std::string& token, const RoomO
       connection_state_ = ConnectionState::Connected;
     }
 
-    // Install listener (Room is fully initialized)
+    // Proof that the issue is fixed
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    LK_LOG_INFO("Room::Connect: sleeping for 1 second");
```
alan-george-lk: Is this the only solution? Do all livekit room connects now get a 1 second delay?

src/tests/integration/test_late_join_track_publication.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+} // namespace
+
+class LateJoinTrackPublicationIntegrationTest
+    : public LiveKitTestBase,
+      public ::testing::WithParamInterface<bool> {};
+
+TEST_P(LateJoinTrackPublicationIntegrationTest,
+       ConsumerReceivesAlreadyPublishedAudioTrackEvents) {
+  skipIfNotConfigured();
```
alan-george-lk: I would avoid using this -- this would cause bad setup tests in CI to just skip and not fail the stage. I think it made sense in a time where skipping tests locally wasn't a big deal, but now that we have it working in CI, assuming it comes from the testing common header, we may want to remove it entirely in a separate PR

src/room.cpp
ladvoc: This gets removed; this was just to validate the fix.

.gitignore
ladvoc: This also gets removed.

src/room.cpp
alan-george-lk: Ok, thanks. The review was requested so was assuming this was part of the fix

.gitignore
stephen-derosa: going to keep this in since its a nice place to put user stories that being reproduced/investigated. I can rename to `debug_workspace` or something else if youd like

src/tests/integration/test_late_join_track_publication.cpp
stephen-derosa: removed

include/livekit/room.h
```diff
@@ -303,6 +303,7 @@ class Room {
 
   mutable std::mutex lock_;
   ConnectionState connection_state_ = ConnectionState::Disconnected;
+  bool connect_in_progress_ = false;
```
stephen-derosa: this is only needed to log when we get events/callback and havent fully init

src/tests/integration/test_late_join_track_publication.cpp
```diff
@@ -0,0 +1,641 @@
...
+  }
+
+  void validateDataCallbackState(Room& room, const std::string& track_name, const std::string& publisher_identity,
+                                 const std::string& callback_name) {
+    validateCommonCallbackState(room, callback_name);
+
+    std::string expected_publisher_identity;
+    bool expected_track = false;
+    {
+      std::lock_guard<std::mutex> lock(state_.mutex);
```
alan-george-lk: FYI, we don't have clang-tidy enabled on tests, but generally it prefers `std::scoped_lock` over `std::lock_guard`: https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-scoped-lock.html

src/tests/stress/test_room_listener_race_stress.cpp
```diff
@@ -0,0 +1,164 @@
...
+  return errors.empty() ? "<none>" : out.str();
+}
+
+} // namespace
+
+class RoomListenerRaceStressTest : public LiveKitTestBase {};
+
+TEST_F(RoomListenerRaceStressTest, ConnectFailDestroyReconnectAndPublishDataTracksConcurrently) {
+  if (!config_.available) {
+    GTEST_SKIP() << "LIVEKIT_URL, LIVEKIT_TOKEN_A, and LIVEKIT_TOKEN_B not set";
```
alan-george-lk: I'd remove this and any future skip, will all be available in CI

xianshijing-lk (approved):
lgtm. 

can you search FfiClient::instance().AddListener() in our code base, and make sure we are using the same patter for all use cases ?

stephen-derosa:
> lgtm.
> 
> can you search FfiClient::instance().AddListener() in our code base, and make sure we are using the same patter for all use cases ?

@xianshijing-lk :
Done. The production call sites are Room, AudioStream, VideoStream, and DataTrackStream.
AudioStream and VideoStream already install their FFI listener before sending the FFI request that creates the stream, then remove it in close()/destruction. DataTrackStream receives an existing subscription handle, then installs the listener before read() starts requesting frames. Room now follows the same event-safety pattern by installing the listener before connectAsync(), and it removes the listener on failed Connect() so reconnect does not leave duplicate callbacks.
I didn’t find any other AddListener() usage that needs a matching change.

src/ffi_client.cpp
```diff
@@ -35,7 +35,6 @@
 namespace livekit {
 
 namespace {
-
```
alan-george-lk, line 38: These are the critical changes I like to see! ;)

src/room.cpp
```diff
@@ -104,16 +104,24 @@ void Room::setDelegate(RoomDelegate* delegate) {
...
     if (connection_state_ != ConnectionState::Disconnected) {
       throw std::runtime_error("already connected");
     }
     connection_state_ = ConnectionState::Reconnecting;
   }
-  auto fut = FfiClient::instance().connectAsync(url, token, options);
+
+  FfiClient::ListenerId listenerId = 0;
   try {
+    // Install listener (Room is fully initialized) [2]
```
alan-george-lk: What does `[2]` and `[1]` mean in this instance? 2 comes before 1 here? maybe clarify this a bit more

alan-george-lk (approved):

src/tests/stress/test_room_listener_race_stress.cpp
alan-george-lk: I just realized this is a stress test. Maybe makes sense to skip, but feels like an error condition to the test IMHO

stephen-derosa: removed GTEST_SKIP

src/room.cpp
stephen-derosa: ty, these were from a debugging session

stephen-derosa: removed

src/tests/integration/test_late_join_track_publication.cpp
stephen-derosa: done

stephen-derosa:
new unit tests shows that the race condition does still exist. We are going to move forward with a "ready" signal from the client sdk to the rust sdk which signals to rust that the client is ready for room events. see this slack thread for discussion: https://live-kit.slack.com/archives/C02EWQYP4KE/p1778708015361339?thread_ts=1778285077.084319&cid=C02EWQYP4KE

src/tests/integration/test_late_join_track_publication.cpp
```diff
@@ -0,0 +1,641 @@
...
+  return failures.empty() ? "<none>" : out.str();
+}
+
+std::string makeTrackName(const std::string& prefix, int index) {
+  return prefix + "-" + std::to_string(index) + "-" + std::to_string(getTimestampUs());
+}
+
+} // namespace
+
+class LateJoinTrackPublicationIntegrationTest : public LiveKitTestBase, public ::testing::WithParamInterface<bool> {};
```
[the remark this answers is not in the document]
stephen-derosa: good suggestion, in our case we want to force interation tests to run, so we throw if not configured

PR #124, opened 2026-05-12
Try protobuf-lite instead

alan-george-lk:
This PR does the following:

- Uses `protobuf-lite` over full `protobuf`. Reduces output library size considerably
  - For `liblivekit.dylib`, before: 5.78 MiB, after: 4.32 MiB, delta: `-25.2%` on MacOS

- Updates documentation to reflect this change


alan-george-lk:
Results not worth the CMake bloat -- closing for now.

PR #125, opened 2026-05-12
FfiClient state guarding, additional testing coverage

alan-george-lk:
This PR does the following:
- Now properly tracks header test coverage for templated types
- Improves general unit test coverage from `15%` --> `31%` (also helps verify symbol visibility)
- Splits/makes more consistent `test_room` into `unit/` and `integration/` folders
- `FfiClient` now checks if the FFI layer is initialized before methods work and throws otherwise
  - This will "fail fast" SDK operations if a user doesn't properly call `livekit::initialize` 

Bug fixes:

- (minor) New visibility controls exposed issue with `FfiClient.h`'s singleton instance conflicting with `livekit/livekit.h`'s instance at the library level (only affected unit testability)


src/tests/integration/test_room.cpp
```diff
@@ -19,78 +19,8 @@
 
 namespace livekit::test {
 
-class RoomTest : public ::testing::Test {
```
alan-george-lk, line 22: FYI for review: these aren't deleted -- moved to a same-named unit test (didn't depend on server). I am starting a convention of the same files/test names across test modalities, i.e.:

- unit: `test_room.cpp`, test case: `RoomTest`
- integration: `test_room.cpp`, test case: `RoomTest`

Since they are entirely different built binaries, no conflicts, and consistency for what's under test from the output.

src/ffi_client.cpp
```diff
@@ -141,6 +141,11 @@ std::optional<FfiClient::AsyncId> ExtractAsyncId(const proto::FfiEvent& event) {
 
 } // namespace
 
+FfiClient& FfiClient::instance() noexcept {
```
alan-george-lk, line 144: Moving this here was critical so that different include-ers of `FfiClient.h` don't get different instances of the singleton.  This came up in the unit test (lib had one, unit test binary had a different one)

src/tests/unit/test_ffi_client.cpp
```diff
@@ -0,0 +1,262 @@
...
+  // contract here so a future refactor that changes it is a deliberate choice.
+  FfiClient::instance().shutdown();
+  EXPECT_NO_THROW(FfiClient::instance().RemoveListener(id));
+}
+
+// ---------------------------------------------------------------------------
+// These tests ensure FfiClient methods throw in various error conditions
+// ---------------------------------------------------------------------------
+
+TEST_F(FfiClientTest, SendRequestThrowsOnEmptyRequest) {
```
alan-george-lk, line 147: The bulk of my work in this branch was ensuring various `FfiClient` operations threw when not initialized. Feel free to discuss if you disagree, but otherwise callbacks don't work when not init'd

src/tests/integration/test_room.cpp
stephen-derosa, line 22: FMU why did we remove these?

stephen-derosa (approved):
looks good to me, one question

src/tests/integration/test_room.cpp
alan-george-lk: I tried to preempt this question but maybe wasn't visible in your review diff: https://github.com/livekit/client-sdk-cpp/pull/125#discussion_r3244819493

stephen-derosa: ty for the cleanup!

include/livekit/livekit.h
```diff
@@ -62,6 +62,9 @@ enum class LogSink {
 ///          already initialized.
 LIVEKIT_API bool initialize(const LogLevel& level = LogLevel::Info, const LogSink& log_sink = LogSink::kConsole);
 
+/// Returns true if the LiveKit SDK is initialized, false otherwise.
+LIVEKIT_API bool isInitialized();
```
xianshijing-lk: do we need such API ? 

can't we just add a DCHECK or assert that it will crash the app if developers call some APIs without calling initialize() ?
I feel like that is what our developers need to do anyway

alan-george-lk: I'm not convinced we need it on the user-side. Happy to remove

src/tests/unit/test_ffi_client.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include "ffi.pb.h"
+#include "ffi_client.h"
+
+namespace livekit::test {
+
+class FfiClientTest : public ::testing::Test {
+protected:
+  void SetUp() override {
+    // Ensure the singleton for this test case starts up uninitialized
+    livekit::shutdown();
```
xianshijing-lk, line 32: I don't really understand why we need such setup. Do different tests share the same singleton? if not, shouldn't singleton starts with uninitialized state by default ?

src/tests/unit/test_ffi_client.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+TEST_F(FfiClientTest, DefaultUninitialized) { EXPECT_FALSE(FfiClient::instance().isInitialized()); }
+
+TEST_F(FfiClientTest, Initialize) {
+  EXPECT_TRUE(FfiClient::instance().initialize(false));
+  EXPECT_TRUE(FfiClient::instance().isInitialized());
+}
+
+TEST_F(FfiClientTest, InitializeFromSDK) {
+  EXPECT_TRUE(livekit::initialize(livekit::LogLevel::Info, livekit::LogSink::kConsole));
```
xianshijing-lk, line 59: this reminds me, do we pass the logs from Rust to C++, do we have some code to test different LogSink ?

src/tests/unit/test_stats.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <gtest/gtest.h>
+#include <livekit/stats.h>
+
+#include <vector>
+
+#include "stats.pb.h"
+
+namespace livekit::test {
+
+TEST(StatsTest, ScalarFromProtoOverloads) {
```
xianshijing-lk, line 26: curiously, do we have public APIs to get the webrtc stats ? like the stacks of the session, track or room ?

src/tests/unit/test_ffi_client.cpp
alan-george-lk: This is defensive programming and maybe isn't needed, but because we're dealing with singletons in a single unit test binary, if a previous test didn't shutdown on cleanup the tests could fail here.

src/tests/unit/test_ffi_client.cpp
alan-george-lk: We do pass logs from Rust to C++ to my knowledge, but I think we're missing a unit test for it. I can add one in this branch.

For the second piece: https://github.com/livekit/client-sdk-cpp/blob/main/src/tests/unit/test_logging.cpp#L72-L75

src/tests/unit/test_stats.cpp
alan-george-lk: We have lots of stuff in `stats.h`, but from a bit of searching around in the repo, I don't think much if any of it is actually exposed in runtime contexts

include/livekit/livekit.h
alan-george-lk: Removed.

src/tests/unit/test_ffi_client.cpp
xianshijing-lk: if multiple tests are sharing the same singleton, the new tests call shutdown(), wouldn't it fail the ongoing old tests ?

src/tests/unit/test_ffi_client.cpp
xianshijing-lk: Thanks

src/tests/unit/test_stats.cpp
xianshijing-lk: yes, the stats.h is just some basic structs that we expose. We will still need the public APIs to expose those runtime stats. Do we have time to do it before the release ?

xianshijing-lk (approved):
lgtm, but I think we need plans to follow-up on some of the questions there.

src/tests/unit/test_stats.cpp
alan-george-lk: Ticket added for that

src/tests/unit/test_ffi_client.cpp
alan-george-lk: It's all synchronous, no tests execute in parallel, but it's still a single process dealing with static memory. So from test-to-test if things were not shut down right from the prior test, that state could leak into the next test for `FfiClient`.

We have a debug assert ensuring the `FfiClient` is shut down in the destructor, but that does not fire on unit/integration tests which are built in release.

I should have addressed your original question more directly, sorry:

> Do different tests share the same singleton?

Yes. Below is the output from two completely separate GTest files

```
[----------] 2 tests from FfiSingletonTest
[ RUN      ] FfiSingletonTest.Test_A
FFI singleton pointer: 0x103697218
[       OK ] FfiSingletonTest.Test_A (0 ms)
[ RUN      ] FfiSingletonTest.Test_B
FFI singleton pointer: 0x103697218
[       OK ] FfiSingletonTest.Test_B (0 ms)
[----------] 2 tests from FfiSingletonTest (0 ms total)
```

PR #126, opened 2026-05-12
fix e2ee defaults masked in release builds

stephen-derosa:
# Overview
correctly populate defaults for e2ee proto

# Testing
debug integration tests running locally

# Linear Ticket
https://linear.app/livekit/issue/BOT-357/e2ee-missing-proto-fields

### DIFF
--- a/include/livekit/e2ee.h
+++ b/include/livekit/e2ee.h
@@ -34,10 +34,18 @@ enum class EncryptionType {
   CUSTOM = 2,
 };
 
-/* Defaults (match other SDKs / Python defaults). */
+/* Key derivation algorithm used by the key provider. */
+enum class KeyDerivationFunction {
+  PBKDF2 = 0,
+  HKDF = 1,
+};
+
+/* Defaults (match Rust KeyProviderOptions::default()). */
 inline constexpr const char* kDefaultRatchetSalt = "LKFrameEncryptionKey";
 inline constexpr int kDefaultRatchetWindowSize = 16;
 inline constexpr int kDefaultFailureTolerance = -1;
+inline constexpr int kDefaultKeyRingSize = 16;
+inline constexpr KeyDerivationFunction kDefaultKeyDerivationFunction = KeyDerivationFunction::PBKDF2;
 
 /**
  * Options for configuring the key provider used by E2EE.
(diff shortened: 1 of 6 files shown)


src/ffi_client.cpp
```diff
@@ -37,6 +37,8 @@ namespace livekit {
 
 namespace {
 
+constexpr int kDefaultKeyRingSize = 16;
```
alan-george-lk: Nit/suggestion: maybe add a comment referencing where this magic default comes from

alan-george-lk (approved):

src/ffi_client.cpp
stephen-derosa: no longer relevant, we use the default from the proto

alan-george-lk (approved):

include/livekit/e2ee.h
```diff
@@ -34,10 +34,18 @@ enum class EncryptionType {
   CUSTOM = 2,
 };
 
-/* Defaults (match other SDKs / Python defaults). */
+/* Key derivation algorithm used by the key provider. */
+enum class KeyDerivationFunction {
+  PBKDF2 = 0,
+  HKDF = 1,
```
xianshijing-lk, line 40: just want to double check,  do we support both ?

include/livekit/e2ee.h
```diff
@@ -34,10 +34,18 @@ enum class EncryptionType {
   CUSTOM = 2,
 };
 
-/* Defaults (match other SDKs / Python defaults). */
+/* Key derivation algorithm used by the key provider. */
+enum class KeyDerivationFunction {
```
xianshijing-lk, line 38: should this call key_derivation_algorithm rather than key_derivation_function ? 
or maybe some other better name ?

xianshijing-lk (approved):
one nit and one question.

lgtm

include/livekit/e2ee.h
stephen-derosa: i will add an integration test for both

stephen-derosa: added in latest commit

include/livekit/e2ee.h
stephen-derosa: i am following the e2ee.proto form of `KeyDerivationFunction`

PR #127, opened 2026-05-13
Unified method style

alan-george-lk:
This PR addresses the following:

- Through an audit of the current state of the repo, the vast majority of functions/methods were lower camel-case styling (AKA `camelBack()`). Some were `PascalCase` and `snake_case`
- **Moving from this PR forward, all functions/methods should be `camelBack()` case**
  - Public headers: methods that violate this rule are duplicated, old ones kept and marked `[[deprecated]]`, such that we have some time to fix other references (examples) prior to release
  - Internal header symbols were updated outright (no duplication/deprecation needed)
  - Unit tests/docs/etc. updated
- `clang-tidy` now enforces this and will fail builds that violate this

> Note: the current plan is to drop the old style deprecated methods for 1.0.0 release. This PR doesn't attempt that just yet

This also addresses a regression in `participant.h`:
- Previously had `set_*` methods (plain member variable setters)
- Conflicted with `local_participant.h` versions of the same name after method unification
- Decision: Now those base methods are removed, and `room.h` is now a friend class and can directly modify those variables. [Matches Python SDK equivalent](https://github.com/livekit/python-sdks/blob/1dfb135952d452f08e315ac98f58b6bec167c4f9/livekit-rtc/livekit/rtc/room.py#L793)


stephen-derosa (approved):
LGTM!

include/livekit/local_participant.h
```diff
@@ -108,14 +108,54 @@ class LIVEKIT_API LocalParticipant : public Participant {
...
+  /**
+   * Update this participant's display name on the server.
+   *
+   * Sends an FFI request to the LiveKit server to change the name
+   * associated with the local participant. Other participants will be
+   * notified via \c onParticipantNameChanged.
+   *
+   * Note: this requires \c canUpdateOwnMetadata permission.
+   */
   void setName(const std::string& name);
```
xianshijing-lk, line 131: curiously, why removing noexcept keyword?

include/livekit/participant.h
```diff
@@ -54,17 +54,63 @@ class LIVEKIT_API Participant {
...
-  void set_kind(ParticipantKind kind) noexcept { kind_ = kind; }
-  void set_disconnect_reason(DisconnectReason reason) noexcept { reason_ = reason; }
+  // ---------------------------------------------------------------------------
+  // Deprecated public mutators
+  // ---------------------------------------------------------------------------
+
+  // NOLINTBEGIN(readability-identifier-naming)
+
+  // Deprecated - see setName() (also deprecated; see notes above).
+  [[deprecated("Participant::set_name is deprecated; use LocalParticipant::setName instead")]]
```
xianshijing-lk, line 64: want to make sure we align on the plan, when will will we remove these deprecated APIs ? 
In the next release ? or after x months ?

xianshijing-lk (commented):
lgtm assuming you address the noexcept

xianshijing-lk (approved):

include/livekit/participant.h
alan-george-lk: Resolved internally

PR #128, opened 2026-05-14
Adjust result.h error condition behavior

alan-george-lk:
This PR addresses a minor API change around `result.h` (dropping `noexcept` label)

Context:
- `Result` accessors were previously marked `noexcept`
- `clang-tidy` flagged this as inaccurate because `std::get` on a `std::variant` can throw a `std::bad_variant_access` if the type being accessed doesn't match the position

Considered alternatives:
- Keeping `noexcept` but using something like `std::get_if` (doesn't throw) --> declined due to dereferencing `nullptr` to maintain API signature
- Switching away from `std::variant` entirely to avoid to avoid this issue --> declined because it still doesn't address the root issue, which is accessing an error result on success and vice-versa
- Adjusting the API to return `std::optional` for accessors --> declined as too impactful to API, and conditionals `ok()` and `has_error()` already exist, would be redundant

Decision:
- `result.h` accessor methods drop `noexcept`
- Throw an exception with a message ourselves vs. going through `std::get` (users can't do much with a generic `std::bad_variant_access` exception with no message)

Impact:

This only affects users of `result.h` that were expecting an exception-free experience, which is negligible if not completely nonexistent at this point in the SDK maturity. In the existing version before this PR, the exception would still actually be thrown if the wrong type was accessed, even though it was marked `noexcept`.

## Testing

- Additional unit tests to cover this flow

### DIFF
--- a/include/livekit/result.h
+++ b/include/livekit/result.h
@@ -16,8 +16,8 @@
 
 #pragma once
 
-#include <cassert>
 #include <optional>
+#include <stdexcept>
 #include <type_traits>
 #include <utility>
 #include <variant>
@@ -34,9 +34,10 @@ namespace livekit {
  * - a success value of type `T`, or
  * - an error value of type `E`
  *
- * Accessors are intentionally non-throwing. Calling `value()` on an error
- * result, or `error()` on a success result, is a programmer error and will
- * trip the debug assertion.
+ * Accessors validate their preconditions before returning. Calling `value()`
+ * on an error result, or `error()` on a success result, throws
+ * `std::runtime_error`. Always check `ok()` / `has_error()` (or the `bool`
+ * conversion) before calling `value()` or `error()`.
  */
 template <typename T, typename E>
 class [[nodiscard]] Result {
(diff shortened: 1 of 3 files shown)


include/livekit/result.h
```diff
@@ -16,8 +16,11 @@
 
 #pragma once
 
+// TODO (AEG) This include (cassert) leaked into downstream cpp examples, removing it here breaks builds.
```
stephen-derosa: dang nice find, are you planning on doing this as a part of this work? If not, consider making a ticket to track

src/tests/unit/test_result.cpp
```diff
@@ -131,6 +132,40 @@ TEST(ResultTest, FailureStringError) {
   EXPECT_EQ(r.error(), "something went wrong");
 }
 
+// ---------------------------------------------------------------------------
```
stephen-derosa, line 135: <3

stephen-derosa (approved):
nice LGTM

include/livekit/result.h
alan-george-lk: I think I'm going to update examples first, then adjust that in this PR, that way there's not 3 total PRs

include/livekit/result.h
```diff
@@ -60,64 +60,83 @@ class [[nodiscard]] Result {
...
   }
 
-  /// Move the success value out. Requires `ok() == true`.
-  // NOLINTNEXTLINE(bugprone-exception-escape)
-  const T&& value() const&& noexcept {
-    assert(ok());
+  /// Move the success value out.
+  ///
+  /// @throws std::logic_error if `ok() == false`.
+  const T&& value() const&& {
```
xianshijing-lk, line 96: Question:
do we such overload ? any existing use case  ? 

I wonder if we should just remove the const T&&  /  const E&& overloads since they do not bring much practical benefit

xianshijing-lk (approved):
One question, lgtm in general

include/livekit/result.h
alan-george-lk: @stephen-derosa any insight here? I agree not a ton of value for such a small data structure. Gonna assume this was agentic thorough-ness and not intentional

stephen-derosa: this was agentic as future proofing -- but i agree i think its over-engineered

include/livekit/result.h
alan-george-lk: Fixed

include/livekit/result.h
alan-george-lk: I dug into this a bit and `E&& error() &&` is actually used by two other classes, so I think we should keep the others in for consistency/avoid API breakage where possible

PR #129, opened 2026-05-14
Fix visibility for RPC

stephen-derosa:
@alan-george-lk summary of what's going on here:

- The RPC integration tests ensure a `livekit::RpcError` was caught in an error-condition test. This error extends `std::runtime_error`
- These tests were not enabled on Mac due to performance issues (speculative) with Mac GHA runners
  - This was missed during the initial visibility PR due to this
- Mac's `.dylib` type info behavior is different than on Linux, which does run and pass these integration tests
- Adjusting the visibility controls to default when not just building the livekit SDK (internal to the SDK build itself, not consumers or unit tests) fixes this visibility issue on the `livekit::RpcError` type so it can be correctly caught (on Mac)

This likely would have popped up in other ways as consumers used the SDK and wanted to inherit from our types (on Mac).

### DIFF
--- a/include/livekit/visibility.h
+++ b/include/livekit/visibility.h
@@ -20,7 +20,8 @@
 //
 // On Unix, the SDK is built with -fvisibility=hidden / -fvisibility-inlines-hidden,
 // so every symbol defaults to hidden. LIVEKIT_API re-exposes the symbol with
-// default visibility. Consumers see no annotation (just a normal declaration).
+// default visibility. Consumers also need default visibility so public RTTI
+// stays unique across shared-library boundaries.
 //
 // On Windows, the SDK is built without WINDOWS_EXPORT_ALL_SYMBOLS, so symbols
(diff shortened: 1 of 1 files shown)


include/livekit/rpc_error.h
```diff
@@ -77,6 +77,7 @@ class LIVEKIT_API RpcError : public std::runtime_error {
    * @param data     Optional extra data, e.g. JSON. Empty string means no data.
    */
   RpcError(ErrorCode code, std::string message, std::string data = {});
+  ~RpcError() override;
```
stephen-derosa: remove if not needed

alan-george-lk (approved):
Approved, thanks. If we run into future issues I think we can verbatim use the one here (with our own LIVEKIT_API labels): https://gcc.gnu.org/wiki/Visibility

But for now this generally matches what we want.

PR #133, opened 2026-05-18
Forward Rust logs through FFI to SDK logger

alan-george-lk:
This PR does the following:

- Forwards all Rust logs via FFI log event (forwards any log macro invocation from Rust code)
  - Critical note: `livekit::initialize(level, log_sink)` is deprecated as of this PR but is still backwards compatible. `log_sink` did not accurately change logging behavior, and `livekit::setLogCallback` is actually the only way. The FFI log API change is independent of callback behavior, but this PR also ensures logs from SDK or FFI layer work the same sink-wise. Overloaded this with `initialize(level)` and marked the other as deprecated
- Adds unit test
- Bumps Rust FFI submodule to recent release version for better log phrasing/behavior

Screenshot 2026-05-19 at 9 38 55 AM

Minor unrelated changes that happened along the way:
- Also implements `panic` event the same way Python does
- Hardened `AGENTS.md` language around format/tidy

## Testing

Added unit test to verify behavior across Error, Warn, Info and Debug levels (Trace not logged in Rust side).

### DIFF
--- a/src/ffi_client.cpp
+++ b/src/ffi_client.cpp
@@ -276,10 +276,12 @@ extern "C" LIVEKIT_INTERNAL_API void LivekitFfiCallback(const uint8_t* buf, size
 
   // Forward any FFI logs to the SDK logger
   if (event.has_logs()) {
+    // Note: explicitly acquiring the logger here to avoid mutex acquires per-message
+    auto logger = detail::getLogger();
     for (const auto& rec : event.logs().records()) {
-      detail::forwardFfiLog(toSpdlogLevel(rec.level()), rec.target(), rec.message());
+      detail::forwardFfiLog(logger, toSpdlogLevel(rec.level()), rec.target(), rec.message());
     }
-    return; // No need to queue the logs
+    return; // No need to queue the log event from here
   }
 
   // We are in a unrecoverable state, terminate the process
--- a/src/ffi_client.cpp
+++ b/src/ffi_client.cpp
@@ -207,7 +207,36 @@ proto::FfiResponse FfiClient::sendRequest(const proto::FfiRequest& request) cons
   return response;
 }
 
+LogLevel fromProtoLogLevel(proto::LogLevel level) {
+  switch (level) {
+    case proto::LOG_ERROR:
+      return LogLevel::Error;
+    case proto::LOG_WARN:
+      return LogLevel::Warn;
+    case proto::LOG_INFO:
+      return LogLevel::Info;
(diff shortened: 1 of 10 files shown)


src/livekit.cpp
```diff
@@ -25,11 +25,9 @@ bool initialize(const LogLevel& level, const LogSink& log_sink) {
   // Initializes logger if singleton instance is not already initialized
   setLogLevel(level);
   auto& ffi_client = FfiClient::instance();
-  return ffi_client.initialize(log_sink == LogSink::kCallback);
+  return ffi_client.initialize(true);
 }
 
-bool isInitialized() { return FfiClient::instance().isInitialized(); }
```
alan-george-lk, line 31: This accidentally made it in from an older PR (no header signature equivalent)

src/livekit.cpp
```diff
@@ -25,11 +25,9 @@ bool initialize(const LogLevel& level, const LogSink& log_sink) {
   // Initializes logger if singleton instance is not already initialized
   setLogLevel(level);
   auto& ffi_client = FfiClient::instance();
-  return ffi_client.initialize(log_sink == LogSink::kCallback);
```
alan-george-lk, line 28: The `log_sink` variable was not working as I suspect it was intended previously. The argument to `FfiClient::initialize` here was to enable FFI logging (forwards all Rust log macros over the FFI callback as events), but `LogSink::kCallback` as a sink type is basically irrelevant to that feature. And passing that value here didn't actually register any callbacks. We should figure out what to do here, options are:

- Breaking change: simply remove the second arg to `livekit::initialize()`
- Just leave this `.initialize(true)` as-is
- Add a third enum value to `LogSink` called `kOff`, and then this arg becomes `.initialize(log_sink != kOff)`, so users have the ability to turn off the FFI logs. But then that brings up the question, we probably should have a cleaner log control. Just turning off FFI logs while others remain doesn't really make sense

src/ffi_client.cpp
```diff
@@ -251,11 +269,29 @@ void FfiClient::PushEvent(const proto::FfiEvent& event) const {
   }
 }
 
-void LivekitFfiCallback(const uint8_t* buf, size_t len) {
+extern "C" LIVEKIT_INTERNAL_API void LivekitFfiCallback(const uint8_t* buf, size_t len) {
```
stephen-derosa: FMU why use `extern "C"` here?

stephen-derosa:
closing until its relevant again

PR #135, opened 2026-05-20
Documentation cleanup: initial pass

alan-george-lk:
This PR does the following:

- Without changing wording, replaces `/* */` Java-style Doxygen syntax with more modern `///` syntax
- Also adds tags (such as `@throws`) so they actually show up in generated docs, and to be more consistent
- Adds `AGENTS.md` section around Doxygen handling for new code
- Found additional old `Connect()` references which are now fixed


stephen-derosa (approved):
nice, looked closely at the AGENTS.md file, looks good to me. All other diffs were skimmed.

PR #136, opened 2026-05-20
CI/CD: Enable integration tests on mac

stephen-derosa:
- remove Low/HighFpsMultiPacket since its covered in benchmarking
- enable integration tests on mac


xianshijing-lk (approved):

src/tests/integration/test_data_track.cpp
```diff
@@ -287,135 +284,9 @@ void runEncryptedDataTrackRoundTrip(KeyDerivationFunction key_derivation_functio
 
 class DataTrackE2ETest : public LiveKitTestBase {};
 
-class DataTrackTransportTest : public DataTrackE2ETest,
-                               public ::testing::WithParamInterface<std::tuple<double, size_t>> {};
-
 class DataTrackKeyDerivationTest : public DataTrackE2ETest,
                                    public ::testing::WithParamInterface<KeyDerivationFunction> {};
 
-TEST_P(DataTrackTransportTest, PublishesAndReceivesFramesEndToEnd) {
```
alan-george-lk, line 296: Instead of fully deleting this, now that @ladvoc has updated it on the Rust side to not be performance-dependent, I think we should adopt this test here instead: https://github.com/livekit/rust-sdks/commit/03803d22db81712251668eaa6c7fdd1c72c67b9a

alan-george-lk (changes requested):

src/tests/integration/test_data_track.cpp
stephen-derosa: this has been implemeted!

.github/workflows/docker-validate.yml
```diff
@@ -11,7 +11,7 @@ permissions:
 
 env:
   # Pinned commit for cpp-example-collection smoke build (https://github.com/livekit-examples/cpp-example-collection)
-  CPP_EXAMPLE_COLLECTION_REF: f231c0c75028d1dcf13edcecd369d030d2c7c8d4
+  CPP_EXAMPLE_COLLECTION_REF: 46083ea4e5d3def8e44a53148c2c7800131efca0
```
alan-george-lk, line 14: I can't tell from the GitHub diff, can you triple check all the hashes match (this file, the other workflow file with this variable, and the submodule)?

.github/workflows/tests.yml
```diff
@@ -63,13 +63,11 @@ jobs:
           - os: macos-26-xlarge
             name: macos-arm64
             build_cmd: ./build.sh release-tests
-            # E2E not possible on GHA Mac runner currently
-            e2e-testing: false
+            e2e-testing: true
```
alan-george-lk, line 66: Huzzah!

src/tests/integration/test_data_track.cpp
```diff
@@ -316,104 +307,65 @@ TEST_P(DataTrackTransportTest, PublishesAndReceivesFramesEndToEnd) {
...
+      DataTrackFrame frame;
+      frame.payload.assign(payload_len, kTransportPayloadValue);
+      while (keep_publishing.load()) {
+        requirePushSuccess(local_track->tryPush(frame), "Failed to push data frame");
+        std::this_thread::sleep_for(50ms);
       }
-
-      ++received_count;
+    } catch (...) {
+      publish_error = std::current_exception();
```
alan-george-lk: Consider `std::async` over a thread/current exception. I implemented it elsewhere in this test to match rust, so you should have an example. It manages this exception catching for you, and generally feels a bit more modern/like Rust

alan-george-lk (approved):

.github/workflows/docker-validate.yml
stephen-derosa: docker-validate.yml, builds.yml, and the submodule are all set to `46083ea4e5d3def8e44a53148c2c7800131efca0`

src/tests/integration/test_data_track.cpp
stephen-derosa: nice, done

PR #137, opened 2026-05-20
Fix missed cpp example collection hash

alan-george-lk:


stephen-derosa (approved):

stephen-derosa:
cherry picked this into https://github.com/livekit/client-sdk-cpp/pull/136. If this lands first, np

PR #138, opened 2026-05-20
Room stats API

alan-george-lk:
This PR exposes a `getStats()` public API method to `room.h` to match Rust.

## API design decisions

TL/DR: matches Rust SDK as 1:1 as possible.

I opted for the signature `getStats()` to match Rust: https://github.com/livekit/rust-sdks/blob/1848e77a32ca2f302a58323a7a21858165d81aad/livekit/src/room/mod.rs#L831 

`std::future` -> `async` equivalent

The struct `SessionStats` also matches Rust: https://github.com/livekit/rust-sdks/blob/1848e77a32ca2f302a58323a7a21858165d81aad/livekit/src/rtc_engine/rtc_session.rs#L112

Stores vectors of `publisher_stats` and `subscriber_stats` to match Rust.

### Error handling

- I opted to match the existing `Track::getStats()` which throws an exception

## Testing

- Added unit test that checks the error path
- Added integration test that does some basic room operations and then obtains the stats


src/room.cpp
```diff
@@ -267,6 +267,21 @@ ConnectionState Room::connectionState() const {
   return connection_state_;
 }
 
+std::future<Result<SessionStats, GetSessionStatsError>> Room::getStats() const {
```
stephen-derosa: should we name `getStatsAsync()` so its clear this is async/matches the ffi call?

stephen-derosa (approved):
nice LGTM!

src/room.cpp
alan-george-lk: Fair point, but see the PR body: I opted for `getStats()` since the Rust public API was `get_stats()`, the doc comment/future return type indicates async

stephen-derosa: the rust call is not async though right since it has the stats locally?

alan-george-lk: Still an async function signature:

```
pub async fn get_stats(&self)
```

Looks it immediately awaits though, so functionally blocking?

I don't know, I'm not a huge fan of putting `async` in the signature directly, prefer comment

stephen-derosa: hm, my preference is to be explicit. I dont think we have other user facing functions which are truly async but at least some precedent has been set within `FfiClient` which uses it in the signature for all relevant functions. your call.

alan-george-lk: I should have cited this earlier, I also modeled after this one: https://github.com/livekit/client-sdk-cpp/blob/main/include/livekit/track.h#L129-L130

```c++
  // Async get stats
  std::future<std::vector<RtcStats>> getStats() const;
```

stephen-derosa (approved):

PR #140, opened 2026-05-27
PlatformAudio

stephen-derosa:
# Overview
Implement PlatformAudio over ffi

## Examples
https://github.com/livekit-examples/cpp-example-collection/compare/sderosa/BOT-377-platform-audio?expand=1

### DIFF
--- a/include/livekit/local_audio_track.h
+++ b/include/livekit/local_audio_track.h
@@ -31,12 +31,15 @@ class OwnedTrack;
 }
 
 class AudioSource;
+class PlatformAudioSource;
 
 /// Represents a user-provided audio track sourced from the local device.
 ///
-///  `LocalAudioTrack` is used to publish microphone audio (or any custom
-///  audio source) to a LiveKit room. It wraps a platform-specific audio
-///  source and exposes simple controls such as `mute()` and `unmute()`.
+/// `LocalAudioTrack` is used to publish microphone audio or any custom audio
+/// source to a LiveKit room. It wraps a platform-specific audio
+/// source and exposes simple controls such as `mute()` and `unmute()`.
+/// Muting a local audio track stops transmitting audio to the room, but
+/// the underlying source may continue capturing depending on platform behavior.
 ///
 ///  Typical usage:
 ///
@@ -52,6 +55,8 @@ class AudioSource;
 ///
 ///  The track name provided during creation is visible to remote
 ///  participants and can be used for debugging or UI display.
+/// @note Thread-safety: Not thread-safe. This is a thin FFI wrapper with no
+/// internal synchronization.
 class LIVEKIT_API LocalAudioTrack : public Track {
 public:
   /// Creates a new local audio track backed by the given `AudioSource`.
@@ -63,30 +68,55 @@ public:
   ///               directly for frame capture.
   ///
   /// @return A shared pointer to the newly constructed `LocalAudioTrack`.
+  /// @throws std::invalid_argument  If \p source is null.
+  /// @throws std::runtime_error  If the FFI request fails.
   static std::shared_ptr<LocalAudioTrack> createLocalAudioTrack(const std::string& name,
                                                                 const std::shared_ptr<AudioSource>& source);
+
+  /// Creates a new local audio track backed by the given `PlatformAudioSource`.
+  ///
+  /// @param name   Human-readable name for the track. This may appear to
+  ///               remote participants and in analytics/debug logs.
+  /// @param source  The platform source that captures microphone audio
+  ///               automatically through WebRTC's Audio Device Module.
+  ///
+  /// @return A shared pointer to the newly constructed `LocalAudioTrack`.
+  /// @throws std::invalid_argument  If \p source is null.
+  /// @throws std::runtime_error  If the FFI request fails.
+  ///
+  /// @note Thread-safety: Not thread-safe. This is a thin FFI wrapper with no
+  /// internal synchronization.
+  static std::shared_ptr<LocalAudioTrack> createLocalAudioTrack(const std::string& name,
+                                                                const std::shared_ptr<PlatformAudioSource>& source);
 
   /// Mutes the audio track.
   ///
   /// A muted track stops sending audio to the room, but the track remains
   /// published and can be unmuted later without renegotiation.
+  ///
+  /// @throws std::runtime_error  If the FFI request fails.
   void mute();
 
-  /// Unmutes the audio track and resumes sending audio to the room.
+  /// Unmute the audio track.
+  ///
+  /// Resumes sending audio to the room.
+  ///
+  /// @throws std::runtime_error  If the FFI request fails.
   void unmute();
 
-  /// Returns a human-readable string representation of the track,
-  /// including its SID and name. Useful for debugging and logging.
+  /// Return a human-readable string representation of the track.
+  ///
+  /// @return String containing the track SID and name.
   std::string toString() const;
 
   /// Returns the publication that owns this track, or nullptr if the track is
   /// not published.
   std::shared_ptr<LocalTrackPublication> publication() const noexcept { return local_publication_; }
 
-  /// Sets the publication that owns this track.
-  /// Note: std::move on a const& silently falls back to a copy, so we assign
-  /// directly. Changing the virtual signature to take by value would enable
-  /// a true move but is an API-breaking change hence left for a future revision.
+  /// Set the publication that owns this track.
+  ///
+  /// @param publication  Publication that owns this track, or nullptr to clear
+  /// the association.
   void setPublication(const std::shared_ptr<LocalTrackPublication>& publication) noexcept override {
     local_publication_ = publication;
   }
--- a/include/livekit/local_audio_track.h
+++ b/include/livekit/local_audio_track.h
@@ -55,7 +55,6 @@ class PlatformAudioSource;
 ///
 ///  The track name provided during creation is visible to remote
 ///  participants and can be used for debugging or UI display.
-/// @note This operation is not thread-safe.
 class LIVEKIT_API LocalAudioTrack : public Track {
 public:
   /// Creates a new local audio track backed by the given `AudioSource`.
--- a/include/livekit/local_audio_track.h
+++ b/include/livekit/local_audio_track.h
@@ -55,8 +55,7 @@ class PlatformAudioSource;
 ///
 ///  The track name provided during creation is visible to remote
 ///  participants and can be used for debugging or UI display.
-/// @note Thread-safety: Not thread-safe. This is a thin FFI wrapper with no
-/// internal synchronization.
+/// @note This operation is not thread-safe.
 class LIVEKIT_API LocalAudioTrack : public Track {
 public:
   /// Creates a new local audio track backed by the given `AudioSource`.
@@ -69,7 +68,7 @@ public:
   ///
   /// @return A shared pointer to the newly constructed `LocalAudioTrack`.
   /// @throws std::invalid_argument  If \p source is null.
-  /// @throws std::runtime_error  If the FFI request fails.
+  /// @throws std::runtime_error If the FFI request fails.
   static std::shared_ptr<LocalAudioTrack> createLocalAudioTrack(const std::string& name,
                                                                 const std::shared_ptr<AudioSource>& source);
 
@@ -82,10 +81,9 @@ public:
   ///
   /// @return A shared pointer to the newly constructed `LocalAudioTrack`.
   /// @throws std::invalid_argument  If \p source is null.
-  /// @throws std::runtime_error  If the FFI request fails.
+  /// @throws std::runtime_error If the FFI request fails.
   ///
-  /// @note Thread-safety: Not thread-safe. This is a thin FFI wrapper with no
-  /// internal synchronization.
+  /// @note This operation is not thread-safe.
   static std::shared_ptr<LocalAudioTrack> createLocalAudioTrack(const std::string& name,
                                                                 const std::shared_ptr<PlatformAudioSource>& source);
 
@@ -94,14 +92,14 @@ public:
   /// A muted track stops sending audio to the room, but the track remains
   /// published and can be unmuted later without renegotiation.
   ///
-  /// @throws std::runtime_error  If the FFI request fails.
+  /// @throws std::runtime_error If the FFI request fails.
   void mute();
 
   /// Unmute the audio track.
   ///
   /// Resumes sending audio to the room.
   ///
-  /// @throws std::runtime_error  If the FFI request fails.
+  /// @throws std::runtime_error If the FFI request fails.
   void unmute();
 
   /// Return a human-readable string representation of the track.
--- a/include/livekit/platform_audio.h
+++ b/include/livekit/platform_audio.h
@@ -171,15 +171,21 @@ public:
   /// @note This operation is not thread-safe.
   PlatformAudio& operator=(PlatformAudio&& other) noexcept = default;
 
-  /// Return the number of recording devices reported when this instance was created.
+  /// Return the current number of recording devices.
+  ///
+  /// @throws PlatformAudioError  If the FFI response is malformed or device
+  /// enumeration fails.
   ///
   /// @note This operation is thread-safe.
-  std::int32_t recordingDeviceCount() const noexcept;
+  std::int32_t recordingDeviceCount() const;
 
-  /// Return the number of playout devices reported when this instance was created.
+  /// Return the current number of playout devices.
+  ///
+  /// @throws PlatformAudioError  If the FFI response is malformed or device
+  /// enumeration fails.
   ///
   /// @note This operation is thread-safe.
-  std::int32_t playoutDeviceCount() const noexcept;
+  std::int32_t playoutDeviceCount() const;
 
   /// Enumerate available microphones.
   ///
--- a/include/livekit/platform_audio.h
+++ b/include/livekit/platform_audio.h
@@ -1,0 +1,269 @@
+/*
+ * Copyright 2026 LiveKit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include "livekit/ffi_handle.h"
+#include "livekit/visibility.h"
+
+namespace livekit {
+
+/// Internal shared state for platform audio handles.
+///
+/// This forward declaration is exposed only so public wrapper types can hold a
+/// shared implementation pointer.
+///
+/// @note Thread-safety: Thread-safe. Instances are managed internally by
+/// PlatformAudio.
+struct PlatformAudioState;
+
+/// Information about a platform audio device.
+///
+/// Device indices may change when audio hardware is added or removed. Prefer
+/// the stable `id` value when selecting a device.
+///
+/// @note Thread-safety: Thread-safe. This is an aggregate value type with no
+/// internal shared state.
+struct AudioDeviceInfo {
+  /// Current device index.
+  std::uint32_t index = 0;
+
+  /// Device name reported by the operating system.
+  std::string name;
+
+  /// Platform-specific stable device identifier.
+  std::string id;
+};
+
+/// Audio processing options for platform microphone capture.
+///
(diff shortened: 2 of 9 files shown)


include/livekit/platform_audio.h
```diff
@@ -0,0 +1,281 @@
...
+/// Audio processing options for platform microphone capture.
+///
+/// The default values enable WebRTC's voice processing path for typical
+/// microphone publishing.
+///
+/// @note Thread-safety: Thread-safe. This is an aggregate value type with no
+/// internal shared state.
+struct PlatformAudioOptions {
+  /// Enable acoustic echo cancellation.
+  bool echo_cancellation = true;
```
stephen-derosa, line 59: these match the rust sdk defaults

include/livekit/platform_audio.h
```diff
@@ -0,0 +1,269 @@
...
+  /// Create an automatically captured microphone source for LocalAudioTrack.
+  ///
+  /// @param options  Audio processing options for the platform microphone path.
+  /// @return Platform-backed audio source suitable for LocalAudioTrack.
+  /// @throws PlatformAudioError  If the FFI response is malformed or source
+  /// creation fails.
+  ///
+  /// @note Thread-safety: Thread-safe. Sends an independent FFI request and
+  /// returns a source that keeps the shared PlatformAudio state alive.
+  std::shared_ptr<PlatformAudioSource> createAudioSource(const PlatformAudioOptions& options = {}) const;
```
xianshijing-lk, line 237: I wonder if getAudioSource is better than createAudioSource ?

Literally, PlatformAudio should have one source as it is a process global webrtc adm.

I think on Rust, we have something like 
platform_audio.rtc_source() ? 
wondering if we should have something similar to rust.

xianshijing-lk (commented):
one question, it looks pretty good from a quick look.

include/livekit/local_audio_track.h
```diff
@@ -52,6 +55,8 @@ class AudioSource;
 ///
 ///  The track name provided during creation is visible to remote
 ///  participants and can be used for debugging or UI display.
+/// @note Thread-safety: Not thread-safe. This is a thin FFI wrapper with no
```
alan-george-lk: These feel a little verbose -- maybe just 

```
/// @note This operation is not thread-safe.
```

and in the inverse

```
/// @note This operation is thread-safe.
```

If you agree let's change and update `AGENTS.md`. If you prefer the verbosity I won't hold anything up over it

include/livekit/local_audio_track.h
```diff
@@ -63,30 +68,55 @@ class LIVEKIT_API LocalAudioTrack : public Track {
...
+  ///
+  /// @throws std::runtime_error  If the FFI request fails.
   void mute();
 
-  /// Unmutes the audio track and resumes sending audio to the room.
+  /// Unmute the audio track.
+  ///
+  /// Resumes sending audio to the room.
+  ///
+  /// @throws std::runtime_error  If the FFI request fails.
```
alan-george-lk: General question about doc comments -- how much does the user need to know about the FFI layer? I wonder if something like this is more clean:

```
/// @throws std::runtime_error If the operation fails.
```

include/livekit/local_audio_track.h
```diff
@@ -63,30 +68,55 @@ class LIVEKIT_API LocalAudioTrack : public Track {
...
+  /// @return String containing the track SID and name.
   std::string toString() const;
 
   /// Returns the publication that owns this track, or nullptr if the track is
   /// not published.
   std::shared_ptr<LocalTrackPublication> publication() const noexcept { return local_publication_; }
 
-  /// Sets the publication that owns this track.
-  /// Note: std::move on a const& silently falls back to a copy, so we assign
-  /// directly. Changing the virtual signature to take by value would enable
```
alan-george-lk, line 88: Thanks for deleting this

include/livekit/platform_audio.h
```diff
@@ -0,0 +1,269 @@
...
+/// This forward declaration is exposed only so public wrapper types can hold a
+/// shared implementation pointer.
+///
+/// @note Thread-safety: Thread-safe. Instances are managed internally by
+/// PlatformAudio.
+struct PlatformAudioState;
+
+/// Information about a platform audio device.
+///
+/// Device indices may change when audio hardware is added or removed. Prefer
```
alan-george-lk: Might be worth a `@warning` or `@note` tag here so it's called out in docs

include/livekit/platform_audio.h
```diff
@@ -0,0 +1,269 @@
...
+/// the stable `id` value when selecting a device.
+///
+/// @note Thread-safety: Thread-safe. This is an aggregate value type with no
```
alan-george-lk: This is a plain struct, doesn't feel like it's needed here

include/livekit/platform_audio.h
```diff
@@ -0,0 +1,269 @@
...
+/// @note Thread-safety: Thread-safe. The source owns an immutable FFI handle
+/// and keeps the shared PlatformAudio state alive.
+class LIVEKIT_API PlatformAudioSource {
+public:
+  /// Copy construction is disabled.
+  ///
+  /// @param other  Source to copy from.
+  ///
+  /// @note Thread-safety: Not thread-safe. This operation is deleted.
+  PlatformAudioSource(const PlatformAudioSource& other) = delete;
```
alan-george-lk, line 87: Nit: can skip doc comments on disabled constructors. Doxygen AFAIK won't even flag that as missing docs.

include/livekit/platform_audio.h
```diff
@@ -0,0 +1,269 @@
...
+
+  PlatformAudioSource(FfiHandle handle, std::shared_ptr<PlatformAudioState> platform_audio) noexcept;
+
+  FfiHandle handle_;
+  std::shared_ptr<PlatformAudioState> platform_audio_;
+};
+
+/// Platform audio device manager backed by WebRTC's Audio Device Module.
+///
+/// Use PlatformAudio for normal microphone publishing when built-in echo
```
alan-george-lk: What is "normal microphone" in this case? Why not just "microphone"

include/livekit/platform_audio.h
alan-george-lk: Agree, but FWIW we tend to use `create...()` in this SDK for other things

src/tests/integration/test_platform_audio.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+class PlatformAudioIntegrationTest : public LiveKitTestBase {};
+
+// Publishing a platform-ADM-backed audio track should reach a remote
+// participant exactly like a manually fed AudioSource track. No real
+// microphone is required: the source captures silence on headless runners,
+// but the publish/subscribe round-trip still completes.
+TEST_F(PlatformAudioIntegrationTest, PublishPlatformAudioTrackEndToEnd) {
+  if (!config_.available) {
+    throw std::runtime_error("LIVEKIT_URL, LIVEKIT_TOKEN_A, and LIVEKIT_TOKEN_B not set");
```
alan-george-lk: I think `FAIL() << "<message>"` is a bit cleaner than throwing, since we're in the GTest world after all

src/tests/integration/test_platform_audio.cpp
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  std::unique_ptr<PlatformAudio> platform_audio;
+  try {
+    platform_audio = std::make_unique<PlatformAudio>();
+  } catch (const PlatformAudioError& error) {
+    GTEST_SKIP() << "PlatformAudio unavailable: " << error.what();
```
alan-george-lk, line 93: Should this be a skip? Per the comment at the top of this test, even without a mic this test should be recording silence, therefore this test can still proceed. This feels like a failure if the constructor throws

src/tests/integration/test_platform_audio.cpp
alan-george-lk: Even cleaner:

```
EXPECT_TRUE(config_.available) << "LIVEKIT URL..."
```

src/tests/integration/test_platform_audio.cpp
```diff
@@ -0,0 +1,239 @@
...
+  std::set<std::string> unsubscribed_sids;
+  std::set<std::string> unpublished_sids;
+};
+
+class PlatformTrackCollectorDelegate : public RoomDelegate {
+public:
+  explicit PlatformTrackCollectorDelegate(PlatformTrackState& state) : state_(state) {}
+
+  void onTrackSubscribed(Room&, const TrackSubscribedEvent& event) override {
+    std::lock_guard<std::mutex> lock(state_.mutex);
```
alan-george-lk, line 50: Feels a bit odd to have the mutex in the state struct vs. just in the root member variable space of this class, is there a design reason for that?

src/tests/integration/test_platform_audio.cpp
```diff
@@ -0,0 +1,239 @@
...
+  ASSERT_NE(track, nullptr);
+
+  TrackPublishOptions publish_options;
+  publish_options.source = TrackSource::SOURCE_MICROPHONE;
+  lockLocalParticipant(*sender_room)->publishTrack(track, publish_options);
+
+  std::unique_lock<std::mutex> lock(receiver_state.mutex);
+  const bool subscribed = receiver_state.cv.wait_for(
+      lock, kSubscriptionTimeout, [&]() { return receiver_state.subscribed_audio_names.count(track_name) > 0; });
+  EXPECT_TRUE(subscribed) << "Receiver never subscribed to the platform audio track";
```
alan-george-lk, line 124: Since we have all this test setup, should we also push the audio frames / verify they are received as well? As far as I understand in the LiveKit terminology, publish/subscribe just advertises/sets up the channels, no payload data makes it across until the push step, right?

The test name is end to end, and would be nice to actually check that part

src/tests/integration/test_platform_audio.cpp
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// Unpublishing a platform audio track must propagate to the remote, exercising
+// the source/track lifecycle that keeps the shared PlatformAudioState alive.
+TEST_F(PlatformAudioIntegrationTest, UnpublishPlatformAudioTrackPropagates) {
+  if (!config_.available) {
+    throw std::runtime_error("LIVEKIT_URL, LIVEKIT_TOKEN_A, and LIVEKIT_TOKEN_B not set");
```
alan-george-lk: Similar cleanup comment to the other test

src/tests/unit/test_platform_audio.cpp
```diff
@@ -0,0 +1,174 @@
...
+  EXPECT_EQ(device.name, "Microphone");
+  EXPECT_EQ(device.id, "device-guid");
+}
+
+TEST_F(PlatformAudioTest, CreateSourceAndTrackWhenAvailable) {
+  std::unique_ptr<PlatformAudio> platform_audio;
+  try {
+    platform_audio = std::make_unique<PlatformAudio>();
+  } catch (const PlatformAudioError& error) {
+    GTEST_SKIP() << "PlatformAudio unavailable: " << error.what();
```
alan-george-lk, line 56: Similar comment here and throughout the test -- I'm skeptical of skipping vs. failing on this. Maybe `EXPECT_NO_THROW(platform_audio = std::make...);`

src/tests/unit/test_platform_audio.cpp
```diff
@@ -0,0 +1,174 @@
...
+  for (const auto& device : playout_devices) {
+    if (!device.id.empty()) {
+      EXPECT_NO_THROW(platform_audio->setPlayoutDevice(device.id));
+      selected_any = true;
+      break;
+    }
+  }
+
+  if (!selected_any) {
+    GTEST_SKIP() << "No audio devices with stable ids available to select";
```
alan-george-lk, line 168: Similar comment here

src/local_audio_track.cpp
```diff
@@ -16,27 +16,58 @@
...
+
+proto::OwnedTrack createAudioTrackWithSourceHandle(const std::string& name, std::uint64_t source_handle) {
+  proto::FfiRequest req;
+  auto* msg = req.mutable_create_audio_track();
+  msg->set_name(name);
+  msg->set_source_handle(source_handle);
+
+  const proto::FfiResponse resp = FfiClient::instance().sendRequest(req);
+  if (!resp.has_create_audio_track()) {
+    // TODO(sderosa): we dont have an error code/return, is throwing ok?
```
alan-george-lk: I think so given the function signature

src/platform_audio.cpp
```diff
@@ -0,0 +1,195 @@
...
+PlatformAudioSource::PlatformAudioSource(FfiHandle handle, std::shared_ptr<PlatformAudioState> platform_audio) noexcept
+    : handle_(std::move(handle)), platform_audio_(std::move(platform_audio)) {}
+
+PlatformAudio::PlatformAudio() {
+  proto::FfiRequest req;
+  req.mutable_new_platform_audio();
+
+  const proto::FfiResponse resp = FfiClient::instance().sendRequest(req);
+  if (!resp.has_new_platform_audio()) {
+    throw PlatformAudioError("FfiResponse missing new_platform_audio");
```
alan-george-lk: Consider adding context here and the other exceptions below, like:

```
"Failed to construct PlatformAudio: <>"
```

That way in console output you have some indicator of where this occurred if you aren't catching/logging it

src/platform_audio.cpp
```diff
@@ -0,0 +1,195 @@
...
+  state_->playout_device_count = owned.info().playout_device_count();
+}
+
+std::int32_t PlatformAudio::recordingDeviceCount() const noexcept {
+  return state_ ? state_->recording_device_count : 0;
+}
+
+std::int32_t PlatformAudio::playoutDeviceCount() const noexcept { return state_ ? state_->playout_device_count : 0; }
+
+std::vector<AudioDeviceInfo> PlatformAudio::recordingDevices() const {
```
alan-george-lk, line 116: Something just came to mind: consider using a verb for operations that aren't plain local accessors (e.g. FFI round trip). For example (hypothetical):

`recordingDevices()` -> just calls into a local `recording_devicies_` vector, no operation

`getRecordingDevices()` -> indicates that there's some sort of exchange into the FFI

We loosely use this pattern elsewhere in the code base, but not uniformly. Might be worth standardizing on in `AGENTS.md`/as a team if you agree, and make those edits here.

I'm not super married to this specific convention, just would be nice to have some indication to the user that "this is a quick lookup of member state" vs. "we're doing a more extensive and error prone operation".

Although if this signature is what the other SDKs use, maybe our hands are tied. I prefer consistency across SDKs vs. deviating for consistency in our code base

src/platform_audio.cpp
```diff
@@ -0,0 +1,195 @@
...
+  proto::FfiRequest req;
+  req.mutable_get_audio_devices()->set_platform_audio_handle(requireHandle(state_));
+
+  const proto::FfiResponse resp = FfiClient::instance().sendRequest(req);
+  if (!resp.has_get_audio_devices()) {
+    throw PlatformAudioError("FfiResponse missing get_audio_devices");
+  }
+
+  const auto& devices = resp.get_audio_devices();
+  if (devices.has_error() && !devices.error().empty()) {
```
alan-george-lk: Is there any situation where an error field is populated without the string body?

src/tests/unit/test_platform_audio.cpp
stephen-derosa: the problem is that our audio devices dont have audio devices

include/livekit/local_audio_track.h
stephen-derosa: good question. We mention the ffi quite a bit in a lot of public API docs. I fear that "if the operation fails" could be misleading to the reason why.

src/tests/integration/test_platform_audio.cpp
stephen-derosa: nice, done

src/tests/integration/test_platform_audio.cpp
stephen-derosa: good point -- changed this to `EXPECT_NO_THROW`

src/tests/unit/test_platform_audio.cpp
stephen-derosa: nvm it still shouldnt throw

src/tests/integration/test_platform_audio.cpp
stephen-derosa: not a bad idea to add a test that sends frames, i can add a new one for that. This test is names `PublishPlatformAudioTrackEndToEnd` -- currently its meant just to test the publishing of the track itself.

include/livekit/platform_audio.h
stephen-derosa: yeah i went with `create...()` to match the rest of the SDK.

If it is the case that the user can not create/get more than one audio source, than i think here `getAudioSource` or even `rtcAudioSource` would make sense

src/tests/integration/test_platform_audio.cpp
stephen-derosa: yeah it doesnt really make sense to have the mutex belong to the delegate, otherwise to access the shared state the calling object (test) would have to reach into the delegate to lock the state.

src/platform_audio.cpp
stephen-derosa: i agree with your point -- but this does in fact match the rust sdk:
```
    pub fn recording_devices(&self) -> impl Iterator<Item = RecordingDeviceInfo> + '_ {
        let count = self.recording_device_count();
        (0..count).filter_map(move |index| self.recording_device_info(index))
    }
```

src/tests/integration/test_platform_audio.cpp
stephen-derosa: added and passes locally -- lets see if its correct for CI

src/platform_audio.cpp
stephen-derosa: yeah the proto is:
```
message GetAudioDevicesResponse {
  // Available playout devices (speakers/headphones).
  repeated AudioDeviceInfo playout_devices = 1;
  // Available recording devices (microphones).
  repeated AudioDeviceInfo recording_devices = 2;
  // Error message if enumeration failed, empty/absent on success.
  optional string error = 3;
}
```

so the `error` field could be present but empty, which indicates a success. So we only throw if the error is present _AND_ its not empty.

src/tests/integration/test_platform_audio.cpp
stephen-derosa: this is actually an incorrect comment -- platform audio requires an audio device, so the `GTEST_SKIP()` is required

src/tests/integration/test_platform_audio.cpp
stephen-derosa: wont run in CI, but it does pass locally

include/livekit/local_audio_track.h
```diff
@@ -52,6 +55,7 @@ class AudioSource;
 ///
 ///  The track name provided during creation is visible to remote
 ///  participants and can be used for debugging or UI display.
+/// @note This operation is not thread-safe.
```
alan-george-lk: Nit: for classes maybe use "This class" instead of operation

alan-george-lk (approved):

include/livekit/platform_audio.h
stephen-derosa: After thinking a bit more I think `createAudioSource()` is the better fit here. `PlatformAudio` reuses the shared platform ADM, but each call creates a new C++ track source handle backed by that ADM, which matches the rest of the SDK's `create...()` sdk namings. I aded a comment to the creteAudioSource function:
```
  /// Each call returns a new track source handle backed by the shared platform
  /// Audio Device Module; it does not create a separate ADM instance.
```

src/platform_audio.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+#include "ffi.pb.h"
+#include "ffi_client.h"
+
+namespace livekit {
+
+struct PlatformAudioState {
+  FfiHandle handle;
+  std::int32_t recording_device_count = 0;
+  std::int32_t playout_device_count = 0;
```
xianshijing-lk: why do we want to cache these |recording_device_count| and |playout_device_count|, they might go out of date if plugging in a new device or unplugging a device

src/platform_audio.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  const auto& owned = platform_audio.platform_audio();
+  state_ = std::make_shared<PlatformAudioState>();
+  state_->handle = FfiHandle(static_cast<uintptr_t>(owned.handle().id()));
+  state_->recording_device_count = owned.info().recording_device_count();
+  state_->playout_device_count = owned.info().playout_device_count();
+}
+
+std::int32_t PlatformAudio::recordingDeviceCount() const noexcept {
+  return state_ ? state_->recording_device_count : 0;
```
xianshijing-lk: I wonder if we should query the device count via ffi request, rather than using the caching values? 

unless rust will update the device count when there is a device changed notification

src/platform_audio.cpp
stephen-derosa: great catch and great point, ty! updated in latest commit.

src/platform_audio.cpp
stephen-derosa: yup absolutely right -- now using `mutable_get_audio_devices`

xianshijing-lk (approved):

PR #143, opened 2026-05-28
Breaking changes: remove deprecated code, return weak_ptr instead of raw, remove unused LogSink enum. More targetted docs.

stephen-derosa:
## Summary
- Remove the deprecated `livekit_bridge` library and all bridge build, CI, Docker, script, docs, and test references.
- Remove previously deprecated C++ SDK API shims and TrackSource-keyed frame callback registration.
- Update subscription callback dispatch so audio/video readers are keyed by `TrackKind`, preventing same-name audio/video tracks from evicting each other.
- Move Doxygen config/assets under `docs/doxygen`, add `scripts/generate-docs.sh`, and add/update docs generation and publish workflows.
- Update unit tests for track-name-only callback registration and callback key collision coverage.
- remove third part warnings
- warnings are now errors

## Internal SDK changes
- Dispatcher event signatures:
  - `SubscriptionThreadDispatcher::handleTrackSubscribed(...)` no longer takes `TrackSource`
  - `SubscriptionThreadDispatcher::handleTrackUnsubscribed(...)` now takes `TrackKind` instead of `TrackSource`

## Public Breaking Changes
- Removed the entire `livekit_bridge` library/API:
  - `liblivekit_bridge` / bridge CMake target
  - `bridge/` source, headers, tests, and README

- Removed `TrackSource`-keyed frame callback APIs:
  - `Room::setOnAudioFrameCallback(participant_identity, TrackSource, ...)`
  - `Room::setOnVideoFrameCallback(participant_identity, TrackSource, ...)`
  - `Room::clearOnAudioFrameCallback(participant_identity, TrackSource)`
  - `Room::clearOnVideoFrameCallback(participant_identity, TrackSource)`
  - Matching `SubscriptionThreadDispatcher` overloads

- Removed deprecated SDK functions:
  - `Room::Connect`
  - `Room::room_info`
  - `AudioFrame::total_samples`, `sample_rate`, `num_channels`, `samples_per_channel`, `to_string`
  - `AudioSource::sample_rate`, `num_channels`, `ffi_handle_id`
  - `Participant::set_name`, `set_metadata`, `set_attributes`, `set_attribute`, `remove_attribute`, `set_kind`, `set_disconnect_reason`
  - `Track::stream_state`, `mime_type`, `has_handle`, `ffi_handle_id`
  - `Result::has_error`
  - `LocalAudioTrack::to_string`
  - `LocalVideoTrack::to_string`
  - `RemoteAudioTrack::to_string`
  - `RemoteVideoTrack::to_string`
  - `RemoteParticipant::to_string`
  - `VideoSource::ffi_handle_id`

- Signature changes:
  - `Room::e2eeManager()` now returns a weak_ptr
  - `Room::localParticipant()` now returns a weak_ptr
  - `Room:remoteParticipant()` now returns a weak_ptr
  - `E2EEManager::keyProvider()` now returns a weak_ptr
  - `livekit::initialize()`removed second argument and the `LogSink` entirely. This was done to future proof releases. This was not decided lightly and was a result of the extensive log forwarding investigation done in #133 and upcoming `v1.0.0` breaking changes anyways. `LogSink` enum was not working as expected. If set to `kCallback`, it would enable FFI `capture_logs`, but no support existed in the FFI client to handle those events. So functionally, `kCallback` would silently disable all Rust logs (never would reach console out or callbacks of any kind). Additionally, setting this value to `kCallback` wouldn't actually forward logs to the user lambda. That only works via `livekit::setLoggingCallback`


include/livekit/room.h
```diff
@@ -176,7 +160,11 @@ class LIVEKIT_API Room {
   RemoteParticipant* remoteParticipant(const std::string& identity) const;
 
   /// Returns a snapshot of all current remote participants.
-  std::vector<std::shared_ptr<RemoteParticipant>> remoteParticipants() const;
+  ///
+  /// @return Vector of non-null pointers to the current remote participants.
+  ///   The pointers are owned by the Room and remain valid until the
+  ///   corresponding participant disconnects.
+  std::vector<RemoteParticipant*> remoteParticipants() const;
```
alan-george-lk: Nice!

src/room.cpp
```diff
@@ -252,12 +246,12 @@ RemoteParticipant* Room::remoteParticipant(const std::string& identity) const {
...
 
-std::vector<std::shared_ptr<RemoteParticipant>> Room::remoteParticipants() const {
+std::vector<RemoteParticipant*> Room::remoteParticipants() const {
   const std::scoped_lock<std::mutex> guard(lock_);
-  std::vector<std::shared_ptr<RemoteParticipant>> out;
+  std::vector<RemoteParticipant*> out;
   out.reserve(remote_participants_.size());
   for (const auto& kv : remote_participants_) {
-    out.push_back(kv.second);
+    out.push_back(kv.second.get());
```
alan-george-lk: Good stuff -- friendly reminder to ensure these pointers are set to `nullptr` during destruction so users can null check before accessing them

README.md
```diff
@@ -621,8 +621,8 @@ lk token create \
 ```
 
 # Deprecation
-- livekit_bridge (bridge/ folder) is deprecated. Avoid using it. Migrate to the base SDK. This will be removed on 06/01/2026.
-- setOn*FrameCallback with TrackSource is deprecated. Use track name instead. This will be removed on 06/01/2026.
+NOTE: With the official 1.0.0 release we have introduced breaking changes to previous unofficial versions in order
```
alan-george-lk: Should we use a quote syntax for this?

> NOTE:

stephen-derosa: done

src/room.cpp
stephen-derosa: done in latest commit here: https://github.com/livekit/client-sdk-cpp/pull/143/changes/ae476e2d96e0dcade01e47aa0bf112d88cafa8f8

src/tests/unit/test_room.cpp
```diff
@@ -33,6 +33,8 @@ TEST_F(RoomTest, ConnectWithoutInitialize) {
   Room room;
   bool result = room.connect("wss://localhost:7880", "test", livekit::RoomOptions());
   EXPECT_FALSE(result) << "Connecting without initializing should return false";
+  EXPECT_EQ(room.localParticipant(), nullptr) << "Local participant should be null after failed connect";
+  EXPECT_TRUE(room.remoteParticipants().empty()) << "Remote participants should be empty after failed connect";
```
alan-george-lk, line 37: Can we also test the participant pointers somewhere, ensure they're null on room destruction? I can add that to this branch if you prefer

stephen-derosa: how would you do that? once the object is destructed how would you access the points? arent they nullptr by default?

or do you mean get the raw pointers, destruct the room, check the pointers?

alan-george-lk: Second one

stephen-derosa: done in latest commit!

include/livekit/room.h
```diff
@@ -148,27 +148,34 @@ class LIVEKIT_API Room {
...
   ///   - identity, SID, metadata
   ///   - publishing/unpublishing operations
-  /// @return Non-null pointer while connected; nullptr before connect, after
-  ///   room end-of-stream teardown, or when the room is destroyed.
-  LocalParticipant* localParticipant() const;
+  ///
+  /// The returned handle is non-owning. Call @c lock() to obtain a usable
+  /// @c weak_ptr; the result is empty (`lock() == nullptr`) before connect,
+  /// after room end-of-stream teardown, or once the room is destroyed. This
+  /// lets callers that cache the handle detect teardown instead of holding a
```
alan-george-lk: I would rephrase "detect tear down" to "detect object lifetime" or something like that, using the lifetime of this object to know if the room is destroyed is a bit backwards, it's just a more explicit contract that the room owns the memory and these are invalid if the room is down

include/livekit/room.h
```diff
@@ -277,15 +287,15 @@ class LIVEKIT_API Room {
   RoomDelegate* delegate_ = nullptr; // Not owned
   RoomInfoData room_info_;
   std::shared_ptr<FfiHandle> room_handle_;
-  std::unique_ptr<LocalParticipant> local_participant_;
+  std::shared_ptr<LocalParticipant> local_participant_;
```
alan-george-lk, line 308: Maybe make a comment that there's only one of these, the `std::shared_ptr` is purely for `std::weak_ptr` support

include/livekit/room.h
```diff
@@ -277,15 +287,15 @@ class LIVEKIT_API Room {
...
   std::unordered_map<std::string, std::shared_ptr<RemoteParticipant>> remote_participants_;
   // Data stream
   std::unordered_map<std::string, TextStreamHandler> text_stream_handlers_;
   std::unordered_map<std::string, ByteStreamHandler> byte_stream_handlers_;
   std::unordered_map<std::string, std::shared_ptr<TextStreamReader>> text_stream_readers_;
   std::unordered_map<std::string, std::shared_ptr<ByteStreamReader>> byte_stream_readers_;
   // E2EE
-  std::unique_ptr<E2EEManager> e2ee_manager_;
+  std::shared_ptr<E2EEManager> e2ee_manager_;
```
alan-george-lk, line 317: Same unique-ness comment here

src/tests/common/test_common.h
```diff
@@ -130,7 +130,7 @@ inline uint64_t getTimestampUs() {
 inline bool waitForParticipant(Room* room, const std::string& identity, std::chrono::milliseconds timeout) {
   auto start = std::chrono::steady_clock::now();
   while (std::chrono::steady_clock::now() - start < timeout) {
-    if (room->remoteParticipant(identity) != nullptr) {
+    if (!room->remoteParticipant(identity).expired()) {
```
alan-george-lk, line 133: Clean!

src/tests/integration/test_data_track.cpp
```diff
@@ -305,9 +307,9 @@ TEST_P(DataTrackTransportTest, PublishesAndReceivesFramesEndToEnd) {
 
   auto rooms = testRooms(room_configs);
   auto& publisher_room = rooms[0];
-  const auto publisher_identity = publisher_room->localParticipant()->identity();
+  const auto publisher_identity = publisher_room->localParticipant().lock()->identity();
 
-  auto local_track = requirePublishedTrack(publisher_room->localParticipant(), track_name);
+  auto local_track = requirePublishedTrack(publisher_room->localParticipant().lock().get(), track_name);
```
alan-george-lk: I think we'd want to change the signature of `requirePublishedTrack` to take the `std::weak_ptr` directly and check (i.e. with an `EXPECT_`) that the pointer isn't expired or wrap in `if (auto ptr = weak_ptr.lock()`, this `.lock().get()` without any guards kinda defeats the point

src/tests/integration/test_data_track.cpp
```diff
@@ -305,9 +307,9 @@ TEST_P(DataTrackTransportTest, PublishesAndReceivesFramesEndToEnd) {
 
   auto rooms = testRooms(room_configs);
   auto& publisher_room = rooms[0];
-  const auto publisher_identity = publisher_room->localParticipant()->identity();
+  const auto publisher_identity = publisher_room->localParticipant().lock()->identity();
```
alan-george-lk: Not going to make a bunch of comments, but in general for this test in others I think we should guard against the `.lock()` first and fail the test otherwise before we access into the data

src/tests/stress/test_rpc_stress.cpp
```diff
@@ -202,6 +202,10 @@ TEST_F(RpcStressTest, MaxPayloadStress) {
   auto start_time = std::chrono::steady_clock::now();
   auto duration = std::chrono::seconds(config_.stress_duration_seconds);
 
+  // Hold the local participant alive for the lifetime of the caller threads.
+  auto caller_lp = caller_room->localParticipant().lock();
+  ASSERT_NE(caller_lp, nullptr);
```
alan-george-lk: I think functionally this works, but it's cleaner/more intended in smart pointers to do:

```
    if (auto caller_lp = caller_room->localParticipant().lock())

   else
       fail() << "Bad localParticipant() access\n";
```

Vs comparing to `nullptr` directly

src/tests/integration/test_data_track.cpp
stephen-derosa: agreed -- its actually unsafe to blindly access the locked pointer without validating

stephen-derosa: created a helper function:
```
inline std::shared_ptr<LocalParticipant> lockLocalParticipant(const Room& room) {
  auto participant = room.localParticipant().lock();
  if (!participant) {
    throw std::runtime_error("Local participant handle is expired");
  }
  return participant;
}
```

to make this cleaner in utests

src/tests/integration/test_data_track.cpp
stephen-derosa: good call -- done

src/tests/stress/test_rpc_stress.cpp
stephen-derosa: agreed -- updated helper `lockRemoteParticipant()` to do this

README.md
```diff
@@ -8,624 +8,235 @@
...
-  --valid-for 99999h \
-  --room robo_room \
-  --grant '{"canPublish":true,"canSubscribe":true,"canPublishData":true}'
-```
+See [docs/tools.md](docs/tools.md).
 
-## Deprecation
+# Deprecation
+> NOTE: With the official 1.0.0 release we have introduced breaking changes to previous unofficial versions in order
+to align with other LiveKit client SDKs. See [PR #143](https://github.com/livekit/client-sdk-cpp/pull/143) for the source code changes.
```
alan-george-lk: Nit: I would avoid referencing PRs from within source code. I like this disclaimer otherwise. Could also add:

```md
- Future deprecations and deprecation dates will be listed here for the next major release
```

.github/workflows/builds.yml
```diff
@@ -58,7 +56,7 @@ env:
   # failing the build.
   SCCACHE_GHA_ENABLED: "true"
   # Pinned commit for cpp-example-collection smoke build (https://github.com/livekit-examples/cpp-example-collection)
-  CPP_EXAMPLE_COLLECTION_REF: 56815733a71c14692569e8adf2916a56a14d4882
+  CPP_EXAMPLE_COLLECTION_REF: 402e6fbcc3cb8b2b2aaf80e21b289f27a9060dc6
```
alan-george-lk, line 59: I get why we have to do this, I would like to rethink this at a later date. Sucks having to reference an inflight branch

alan-george-lk (approved):

PR #145, opened 2026-05-29
Remove second arg to init/LogSink

alan-george-lk:
This PR introduces a breaking change to `livekit::initialize`, removing the second argument, and removes the `LogSink` enum entirely. This was not decided lightly and was a result of the extensive log forwarding investigation done in #133 and upcoming `v1.0.0` breaking changes anyways.

- `LogSink` enum was not working as expected
  - If set to `kCallback`, it would enable FFI `capture_logs`, but no support existed in the FFI client to handle those events. So functionally, `kCallback` would silently disable all Rust logs (never would reach console out or callbacks of any kind)
  - Additionally, setting this value to `kCallback` wouldn't actually forward logs to the user lambda. That only works via `livekit::setLoggingCallback`


docs/doxygen/index.md
stephen-derosa, line 1: FMU why is this no longer needed?

stephen-derosa (approved):

docs/doxygen/index.md
alan-george-lk: Missed this in the Doxygen PR so added here, TLDR this was the home page for Doxygen, but now we're embedding the README as the home page like Python does it: https://docs.livekit.io/reference/client-sdk-cpp/

stephen-derosa:
consumed by https://github.com/livekit/client-sdk-cpp/pull/143

PR #146, opened 2026-05-29
Room disconnect issue fix

### LINKED ISSUE #118: Room destructor does not disconnect from server; shutdown() must be called first (undocumented requirement)
## Summary
Destroying a Room object (via room.reset() or end of scope) does not send a disconnect signal to the LiveKit server.

## Root cause
`~Room()` calls `~FfiHandle()` → `livekit_ffi_drop_handle()`, which only removes the handle from `FFI_SERVER.ffi_handles`. It does not close the room.

`livekit_ffi_dispose()` (called by `livekit::shutdown()`) is what actually iterates `ffi_handles` and calls `room.close(ClientInitiated)` — but only for handles still in the map at that point.

So the following common pattern silently fails to disconnect:
```c++
room.reset();       // drops handle → removed from ffi_handles
livekit::shutdown(); // dispose() finds no rooms → never sends disconnect
```

## Correct Order with current implementation
```c++
livekit::shutdown(); // dispose() finds room in ffi_handles → room.close() → server notified
room.reset();        // C++ wrapper destroyed; FFI handle already gone
```
This is neither documented nor enforced, and contradicts natural RAII expectations (object destruction is sufficient for cleanup).

## Expected behavior
One of:
- **Option A:** `~Room()` sends an explicit `DisconnectRequest` before dropping the FFI handle, making destruction sufficient for a clean leave.
- **Option B:** A public `Room::disconnect()` method is added with documentation requiring it before destruction.
- **Option C:** Document the required `shutdown()`-before-`reset()` ordering in the quickstart guide and in room.h if possible.

alan-george-lk:
This PR does the following:

- Fixes #118 issue reported by user: `~Room` now gracefully disconnects from the server
- Also implements `Room::disconnect()` API call mirroring Python SDK implementation

## Testing

- Added unit and integration tests verifying behavior and replicating original user-reported issue through a `room.reset()` smart pointer invoked destruction

### DIFF
--- a/include/livekit/room.h
+++ b/include/livekit/room.h
@@ -143,22 +143,18 @@ public:
 
   /// Disconnect from the room.
   ///
-  /// Sends a `DisconnectRequest` to the server with the given reason, blocks
-  /// until the FFI acknowledges, and tears down all room state (participants,
-  /// listener, E2EE manager, subscription threads). The `onDisconnected`
-  /// delegate is invoked once with the supplied reason.
+  /// This method attempts a best-effort graceful disconnect of the room. If the room was connected prior, after @ref
+  /// disconnect() is called the room object is considered in a terminal state and should no longer be used. If @ref
+  /// disconnect() was called before @ref connect(), no operations are performed and the room object is still valid.
   ///
+  /// @note `~Room()` invokes `disconnect()` automatically if the room is
+  ///       still connected, so explicit calls are optional.
   /// @warning Safe to call from any thread, but **must not** be called from inside a
   /// `RoomDelegate` callback — doing so will deadlock the event listener.
   ///
-  /// @note `~Room()` invokes `disconnect()` automatically if the room is
-  ///       still connected, so explicit calls are optional. Calling this
-  ///       explicitly lets you handle the disconnect outcome and choose a
-  ///       reason other than `ClientInitiated`.
-  ///
   /// @param reason  Reason reported to the server (default: ClientInitiated).
-  /// @returns true if a disconnect roundtrip was performed; false if the
-  ///          room was already disconnected.
+  /// @returns true if the graceful disconnect succeeds; false if the
+  ///          room was already disconnected (no-op) or the graceful disconnect fails.
   bool disconnect(DisconnectReason reason = DisconnectReason::ClientInitiated);
 
   // Accessors
--- a/include/livekit/room.h
+++ b/include/livekit/room.h
@@ -141,6 +141,26 @@ public:
   bool Connect(const std::string& url, const std::string& token, const RoomOptions& options);
   // NOLINTEND(readability-identifier-naming)
 
+  /// Gracefully disconnect from the room.
+  ///
+  /// Sends a `DisconnectRequest` to the server with the given reason, blocks
+  /// until the FFI acknowledges, and tears down all room state (participants,
+  /// listener, E2EE manager, subscription threads). The `onDisconnected`
+  /// delegate is invoked once with the supplied reason.
+  ///
+  /// Safe to call from any thread, but **must not** be called from inside a
+  /// `RoomDelegate` callback — doing so will deadlock the event listener.
+  ///
+  /// @note `~Room()` invokes `disconnect()` automatically if the room is
+  ///       still connected, so explicit calls are optional. Calling this
+  ///       explicitly lets you handle the disconnect outcome and choose a
+  ///       reason other than `ClientInitiated`.
+  ///
+  /// @param reason  Reason reported to the server (default: ClientInitiated).
+  /// @returns true if a disconnect roundtrip was performed; false if the
+  ///          room was already disconnected.
+  bool disconnect(DisconnectReason reason = DisconnectReason::ClientInitiated);
+
   // Accessors
 
   /// Retrieve static metadata about the room.
--- a/include/livekit/room.h
+++ b/include/livekit/room.h
@@ -141,14 +141,14 @@ public:
   bool Connect(const std::string& url, const std::string& token, const RoomOptions& options);
   // NOLINTEND(readability-identifier-naming)
 
-  /// Gracefully disconnect from the room.
+  /// Disconnect from the room.
   ///
   /// Sends a `DisconnectRequest` to the server with the given reason, blocks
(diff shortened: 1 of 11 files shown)


include/livekit/room.h
```diff
@@ -141,6 +141,26 @@ class LIVEKIT_API Room {
...
   // NOLINTEND(readability-identifier-naming)
 
+  /// Gracefully disconnect from the room.
+  ///
+  /// Sends a `DisconnectRequest` to the server with the given reason, blocks
+  /// until the FFI acknowledges, and tears down all room state (participants,
+  /// listener, E2EE manager, subscription threads). The `onDisconnected`
+  /// delegate is invoked once with the supplied reason.
+  ///
+  /// Safe to call from any thread, but **must not** be called from inside a
```
stephen-derosa: consider making this `@warning`

src/tests/integration/test_room.cpp
stephen-derosa, line 1: nice!

src/ffi_client.h
```diff
@@ -94,6 +95,10 @@ class LIVEKIT_INTERNAL_API FfiClient {
   std::future<proto::ConnectCallback> connectAsync(const std::string& url, const std::string& token,
                                                    const RoomOptions& options);
 
+  // Send a DisconnectRequest for the given room handle and wait for the
```
stephen-derosa: nit use proper doxygen

src/local_participant.cpp
```diff
@@ -439,7 +440,17 @@ void LocalParticipant::handleRpcMethodInvocation(uint64_t invocation_id, const s
   if (response_payload.has_value()) {
     msg->set_payload(*response_payload);
   }
-  FfiClient::instance().sendRequest(req);
+  // This is invoked on the FFI listener thread; an unhandled throw here would
```
stephen-derosa: FMU why is this being added here?

src/room.cpp
```diff
@@ -76,6 +76,21 @@ void readyForRoomEvent(std::uint64_t room_handle) {
 Room::Room() : subscription_thread_dispatcher_(std::make_unique<SubscriptionThreadDispatcher>()) {}
 
 Room::~Room() {
+  // Issue a graceful disconnect so the server sees us leave instead of
+  // timing out (RAII expectation; see issue #118). disconnect() does the
```
stephen-derosa: is this actually 118? https://linear.app/livekit/issue/BOT-118/integrate-all-dependencies-into-core-livekit-component

src/room.cpp
```diff
@@ -234,6 +243,83 @@ bool Room::Connect(const std::string& url, const std::string& token, const RoomO
...
+
+  // Tell the FFI to close the room and wait for the callback. Catch the
+  // exception so we still run teardown below; the caller learns about the
+  // failure via the returned bool / logs.
+  bool ffi_ok = true;
+  if (handle) {
+    try {
+      FfiClient::instance().disconnectAsync(handle->get(), reason).get();
+    } catch (const std::exception& e) {
+      LK_LOG_ERROR("Room::disconnect: FFI disconnect failed: {}", e.what());
```
stephen-derosa: in this case should we be setting `connection_state_` back to the previous state?

include/livekit/room.h
alan-george-lk: Done

src/ffi_client.h
alan-george-lk: Not user-visible, but will do for consistency

src/local_participant.cpp
alan-george-lk: This was a work around to avoid an integration test crash, I think this is a bad idea, still investigating

src/room.cpp
alan-george-lk: GitHub issue #118, I am gonna drop this comment/references to that issue

src/room.cpp
alan-george-lk: Great catch -- fixing

src/local_participant.cpp
alan-george-lk: Removed this and identified the issue

src/ffi_client.cpp
```diff
@@ -380,6 +380,37 @@ std::future<proto::ConnectCallback> FfiClient::connectAsync(const std::string& u
...
+  disconnect->set_request_async_id(async_id);
+  disconnect->set_reason(toProto(reason));
+
+  try {
+    const proto::FfiResponse resp = sendRequest(req);
+    if (!resp.has_disconnect()) {
+      logAndThrow("FfiResponse missing disconnect");
+    }
+  } catch (...) {
+    cancelPendingByAsyncId(async_id);
```
stephen-derosa, line 407: LK_LOG_ERROR ?

src/room.cpp
```diff
@@ -234,6 +219,103 @@ bool Room::Connect(const std::string& url, const std::string& token, const RoomO
...
+  std::shared_ptr<FfiHandle> handle;
+  RoomDelegate* delegate_snapshot = nullptr;
+  std::unique_ptr<LocalParticipant> local_participant_to_cleanup;
+  std::unordered_map<std::string, std::shared_ptr<RemoteParticipant>> remote_participants_to_clear;
+  std::unique_ptr<E2EEManager> e2ee_manager_to_clear;
+  std::unordered_map<std::string, std::shared_ptr<TextStreamReader>> text_stream_readers_to_clear;
+  std::unordered_map<std::string, std::shared_ptr<ByteStreamReader>> byte_stream_readers_to_clear;
+  int listener_to_remove = 0;
+
+  {
```
stephen-derosa, line 231: discussed in slack -- if the sdk [is killed without calling disconnect the](https://docs.livekit.io/intro/basics/connect/#disconnect) sfu drops the connection in 15 seconds, so this is really best effort.

stephen-derosa (approved):
looks good to me -- synced with @ladvoc on this and the order of destroying all objects, then sending an async disconnect request is bang on

src/ffi_client.cpp
alan-george-lk: Reasonable but other operations in this file don't do that pattern in this catch, the above try block will log if that part goes wrong

ladvoc (approved):

PR #147, opened 2026-06-01
ISSUE_TEMPLATE

stephen-derosa:
# Overview
Require certain fields in order to submit a ticket. This will help reduce back and forth on Q/A to reproduce/fix issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,81 @@
+name: Bug Report
+description: File a bug report.
+labels: [triage]
+assignees: [alan-george-lk, stephen-derosa]
+body:
+  - type: input
+    id: sdk-version
+    validations: { required: true }
+    attributes:
+      label: LiveKit C++ SDK Version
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,81 @@
+name: Bug Report
+description: File a bug report.
+labels: [triage]
+assignees: [alan-george-lk, stephen-derosa]
```
alan-george-lk, line 4: Is it possible to reference CODEOWNERS here so this is less hard-coded?

alan-george-lk (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
stephen-derosa: good idea let me see

stephen-derosa: short answer is not through the config.

Long answer: according to opus, we could add a GHA to run on "issue open" which checks CODEOWNERS and assigns based off that.

IMO juice isnt worth the squeeze there. who knows, might might ending up assigning to copilot at some point 😅

PR #150, opened 2026-06-01
Unify clang-tidy args/GitHub summary output with clang-format

alan-george-lk:
This PR does the following:

- Allow developer to pass in file path to `./scripts/clang-tidy.sh <file>` just like `clang-format.sh` to analyze a single file
- Add `--fix` arg to match `clang-format.sh`
- Updated GitHub summary view to look more similar on success

Screenshot 2026-06-01 at 4 11 33 PM

### DIFF
--- a/scripts/clang-tidy.sh
+++ b/scripts/clang-tidy.sh
@@ -69,7 +69,7 @@ repo's CI workflow uses. Picks up checks from the repo-root .clang-tidy.
 Options:
   -h, --help, -?
         Show this help and exit.
-  --fix, -i
+  --fix
         Apply fixes in place (forwarded to run-clang-tidy as -fix).
   --github-actions, --gh
         Force GitHub Actions annotation + step-summary mode.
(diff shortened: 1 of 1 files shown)


scripts/clang-tidy.sh
```diff
@@ -127,13 +137,21 @@ fi
...
 forward_args=()
+explicit_files=()
 while (($#)); do
   case "$1" in
     -h|--help|-\?)
       usage
       __tidy_hint_active=0
       exit 0
       ;;
+    --fix|-fix|-i)
```
stephen-derosa: curious the choice of `--fix`, `-fix` and `-i` -- why not just `--fix`?

stephen-derosa (approved):

scripts/clang-tidy.sh
alan-george-lk: Fixed, was thinking of keeping the old `-fix` and `-i` is a common "input file" short hand, but not super useful here.

PR #151, opened 2026-06-02
RPC V2 Tests

stephen-derosa:
# Overview
RPC V2 supports up to 15KB rpc calls. This integrates the happy path of the [client-sdk-js RPC Specs](https://github.com/livekit/client-sdk-js/blob/00e93de0c3d85f846fa6707749fa46457a653a29/RPC_SPEC.md#minimum-required-test-cases)

These tests explicitly test payloads larger than 15kb

### DIFF
--- a/src/tests/integration/test_rpc.cpp
+++ b/src/tests/integration/test_rpc.cpp
@@ -33,8 +33,9 @@ namespace livekit::test {
 
 using namespace std::chrono_literals;
 
-// Maximum RPC payload size (15KB)
-constexpr size_t kMaxRpcPayloadSize = 15 * 1024;
+// RPC v1 packet payload limit was ~15 KiB; these tests validate that RPC v2 can exceed it.
+constexpr size_t kRpcV1PayloadLimit = 15 * 1024;
+constexpr size_t kLargeRpcPayloadSize = 20 * 1000;
 
 // Test configuration from environment variables
 struct RpcTestConfig {
@@ -90,6 +91,14 @@ std::string generateRandomPayload(size_t size) {
   return result.substr(0, size);
 }
 
+size_t checksumPayload(const std::string& payload) {
+  size_t checksum = 0;
+  for (char c : payload) {
+    checksum += static_cast<unsigned char>(c);
+  }
+  return checksum;
+}
+
 class RpcIntegrationTest : public ::testing::Test {
 protected:
   void SetUp() override {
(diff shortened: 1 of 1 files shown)


src/tests/integration/test_rpc.cpp
```diff
@@ -90,6 +91,14 @@ std::string generateRandomPayload(size_t size) {
   return result.substr(0, size);
 }
 
+size_t checksumPayload(const std::string& payload) {
```
alan-george-lk, line 94: Little confused on the utility of this, is it just to help identify payloads in the `echo:` line below? Why not something like `std::hash`

src/tests/integration/test_rpc.cpp
```diff
@@ -148,23 +166,69 @@ TEST_F(RpcIntegrationTest, BasicRpcRoundTrip) {
...
   lockLocalParticipant(*receiver_room)->unregisterRpcMethod("echo");
   caller_room.reset();
   receiver_room.reset();
 }
 
-// Test maximum payload size (15KB)
-TEST_F(RpcIntegrationTest, MaxPayloadSize) {
+TEST_F(RpcIntegrationTest, LargeRequestPayloadRoundTrip) {
+  if (!config_.available) {
+    GTEST_SKIP() << "LIVEKIT_URL, LIVEKIT_TOKEN_A, and LIVEKIT_TOKEN_B not set";
```
alan-george-lk: I would fail here instead of skip, would pass in CI otherwise

```
EXPECT_TRUE(config_.available) << "Message"
```

Suggest this for all tests in this PR

src/tests/integration/test_rpc.cpp
```diff
@@ -148,23 +166,69 @@ TEST_F(RpcIntegrationTest, BasicRpcRoundTrip) {
...
+  }
+
+  auto receiver_room = std::make_unique<Room>();
+  RoomOptions options;
+  options.auto_subscribe = true;
+
+  bool receiver_connected = receiver_room->connect(config_.url, config_.token_b, options);
+  ASSERT_TRUE(receiver_connected) << "Receiver failed to connect";
```
alan-george-lk: Could combine these into a single line, doesn't look like `receiver_connected` is used anywhere else. Suggest this for all tests in this PR

src/tests/integration/test_rpc.cpp
stephen-derosa: thanks yeah we should definitely fail on missing config, i had been throwing her but failing is better.

src/tests/integration/test_rpc.cpp
stephen-derosa: > is it just to help identify payloads in the echo: line below?
yeah this is the driving factor. One other reason is its easier to debug size of a payload (i.e. see missing bytes) than a hash.

also size is more deterministic than hash across libstdc++/libc++/MSVC. (not _really_ relevant since its on the same machine in a single instance) but if we ever wanted to bring this cross machines it might not work

src/tests/integration/test_rpc.cpp
```diff
@@ -104,40 +113,46 @@ class RpcIntegrationTest : public ::testing::Test {
 
 // Test basic RPC round-trip
 TEST_F(RpcIntegrationTest, BasicRpcRoundTrip) {
-  if (!config_.available) {
-    GTEST_SKIP() << "LIVEKIT_URL, LIVEKIT_TOKEN_A, and LIVEKIT_TOKEN_B not set";
-  }
+  EXPECT_TRUE(config_.available) << "Missing integration configuration";
+  ;
```
alan-george-lk: nit: extra semi colon survived the refactor

alan-george-lk (approved):

src/tests/integration/test_rpc.cpp
stephen-derosa: surprised co-pilot didnt catch

PR #152, opened 2026-06-03
FFI panic event support

alan-george-lk:
- Implements the FFI Panic event, matching the [Python SDK behavior](https://github.com/livekit/python-sdks/blob/be92457e3a6e749e19c3fe35ad87784f4cee3c2d/livekit-rtc/livekit/rtc/_ffi_client.py#L184-L188) raising a `SIGTERM`
- Renamed FFI event callback method to match current styling and Python signature (internal API change, no breakage)
- Updated docs

## Testing

- Added unit test to verify users can catch the `SIGTERM`/generally verify behavior


stephen-derosa (approved):
nice -- lgtm!

PR #158, opened 2026-06-08
Use LiveKit server GitHub action

alan-george-lk:
Refactors out old manually installation/running of LiveKit server for CI integration tests to use new `dev-server-action` instead: https://github.com/livekit/dev-server-action

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -114,7 +114,8 @@ jobs:
             libssl-dev \
             libprotobuf-dev protobuf-compiler \
             libabsl-dev \
-            libwayland-dev libdecor-0-dev
+            libwayland-dev libdecor-0-dev \
+            jq
 
       - name: Install deps (macOS)
         if: runner.os == 'macOS'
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -207,47 +206,15 @@ jobs:
...
-          done
-          echo "::error::livekit-server failed to start within 30s"
-          tail -n 200 livekit-server.log || true
-          exit 1
+        id: livekit_server
+        uses: livekit/dev-server-action@61e2b4dcb170dd3591e0c9b0db3c3fe5db93b500
+        with:
+          github-token: ${{ github.token }}
+          config: |
+            enable_data_tracks: true
```
ladvoc: This is enabled by default now, so the explicit config can be removed.

ladvoc (approved):

.github/workflows/tests.yml
alan-george-lk: Will wait for CI to make sure it works, then fix on next commit. Good to know

stephen-derosa (approved):
nice, LGTM

PR #159, opened 2026-06-09
update rust hash for newer dashmap version, GHA to validate rust hash is released

stephen-derosa:
- [X] requires https://github.com/livekit/rust-sdks/pull/1139/changes to be merged

1. update rust hash
2. GHA to validate rust hash is released

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
       tests:  ${{ steps.filter.outputs.tests }}
       docs:   ${{ steps.filter.outputs.docs }}
       cpp_checks: ${{ steps.filter.outputs.cpp_checks }}
+      rust_submodule: ${{ steps.filter.outputs.rust_submodule }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d  # v4.0.1
@@ -83,6 +84,8 @@ jobs:
               - .clang-tidy
               - .github/workflows/ci.yml
               - .github/workflows/cpp-checks.yml
+            rust_submodule:
+              - client-sdk-rust
 
   builds:
     name: Builds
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       tests:  ${{ steps.filter.outputs.tests }}
       docs:   ${{ steps.filter.outputs.docs }}
       cpp_checks: ${{ steps.filter.outputs.cpp_checks }}
+      rust_submodule: ${{ steps.filter.outputs.rust_submodule }}
```
alan-george-lk: Nit: would prefer something like `rust_release` to better match the workflow name

.github/workflows/rust-release-check.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RUST_REPO: livekit/client-sdk-rust
+          RUST_SHA: ${{ steps.rust_sha.outputs.sha }}
+        run: |
+          set -euo pipefail
+          echo "Pinned client-sdk-rust commit: ${RUST_SHA}"
+
+          # Published (non-draft) releases -> their tag names.
+          gh api --paginate "repos/${RUST_REPO}/releases" \
```
alan-george-lk, line 40: This section works, but requires a few `gh` API calls and parses. I think it could be simpler using a simple `git describe`

```
if git describe --exact-match --tags "$sha" >/dev/null 2>&1; then
  echo "tagged"
else
  echo "not tagged"
fi
```

Now you could argue a tag != released, but I think a tag is safe enough for what I was envisioning. Just making sure it's a known version vs. any point in main. Up to you how you wanna proceed

alan-george-lk (approved):

.github/workflows/ci.yml
stephen-derosa: went with `rust_release_check`

.github/workflows/rust-release-check.yml
stephen-derosa: Im going to keep the current logic because it verifies the commit has a tag AND that tag corresponds to a non-draft GitHub Release. I think the strict check of "is this a tag" is an edge case, but still id like to guard against it.

PR #160, opened 2026-06-10
Harden AGENTS.md around clang usage, cleanup clang-tidy.sh wrapper

alan-george-lk:
I've noticed different agents/models don't always run clang-format/clang-tidy, as the existing language was pretty loose ("if needed"). Adds some more strict rules around it to avoid CI waste.

Also removes the "forward to clang tidy" feature of the script, it's not useful in my experience and made the code more complicated when checking file paths on individual file runs. A file not existing would not be caught prior to this.


stephen-derosa (approved):
LGTM -- if possible try to reduce AGENTS.md content to minimum "required" instructions. Im not sure what the "required" amount is though

PR #161, opened 2026-06-10
Add remaining available RoomOptions fields

alan-george-lk:
- Adds support for missing room options (`RoomOptions` struct)
- Remove adhoc logic to serialize certain room options into the FFI; now just has a `toProto` helper for the entire struct
- Unit tests

### DIFF
--- a/include/livekit/room.h
+++ b/include/livekit/room.h
@@ -76,8 +76,17 @@ struct RoomOptions {
   ///   - remote audio/video frames
   bool auto_subscribe = true;
 
-  /// Enable adaptive stream so the server optimizes subscribed video layers.
-  bool adaptive_stream = false;
+  /// Enable adaptive stream for subscribed video tracks.
+  ///
+  /// When enabled, the SDK tells the server it may adjust the video layers sent
+  /// to this client based on what the application is currently rendering. This
+  /// lets the server pause or downscale subscribed video that is off-screen,
+  /// hidden, or only needed at a smaller size, reducing downstream bandwidth and
+  /// decode work. This affects media received by this room; use @ref dynacast
+  /// to control how this client publishes layers to others.
+  ///
+  /// If unset, the Rust SDK default is used.
+  std::optional<bool> adaptive_stream;
 
   /// Enable dynacast (server sends optimal layers depending on subscribers).
   bool dynacast = false;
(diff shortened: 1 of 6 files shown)


stephen-derosa (approved):
nice -- LGTM!

include/livekit/room.h
```diff
@@ -74,19 +76,28 @@ struct RoomOptions {
...
   bool dynacast = false;
 
+  /// Optional end-to-end encryption settings.
+  std::optional<E2EEOptions> encryption;
+
+  /// Optional WebRTC configuration (ICE policy, servers, etc.)
+  std::optional<RtcConfig> rtc_config;
+
+  /// Number of retries for the initial room join after the first attempt.
+  std::uint32_t join_retries = 3;
```
ladvoc: question: I wonder if this should delegate to Rust for default values? Advantage of this approach is it is visible to the C++ dev, disadvantage is C++ could have different defaults if this later changes on the Rust side.

alan-george-lk: Changed

include/livekit/room.h
```diff
@@ -74,19 +76,41 @@ struct RoomOptions {
...
+  ///
+  /// When enabled, the SDK tells the server it may adjust the video layers sent
+  /// to this client based on what the application is currently rendering. This
+  /// lets the server pause or downscale subscribed video that is off-screen,
+  /// hidden, or only needed at a smaller size, reducing downstream bandwidth and
+  /// decode work. This affects media received by this room; use @ref dynacast
+  /// to control how this client publishes layers to others.
+  ///
+  /// If unset, the Rust SDK default is used.
+  std::optional<bool> adaptive_stream;
```
stephen-derosa, line 93: Question: to what extent should we just reference the official livekit docs?
https://docs.livekit.io/transport/media/subscribe/#adaptive-stream

include/livekit/room.h
```diff
@@ -74,19 +76,41 @@ struct RoomOptions {
...
+
```
stephen-derosa, line 94: went to check what the python sdk has for this fields, but it looks like its not present:
https://github.com/livekit/python-sdks/blob/main/livekit-rtc/livekit/rtc/room.py#L110

stephen-derosa (approved):
looks good to me, not a huge public interface change as it just extends the existing room options

include/livekit/room.h
alan-george-lk: Great idea, I'm happy to do this, but we should also add a link-check CI stage like @ladvoc did for ESP32. I could whip that up real quick

include/livekit/room.h
alan-george-lk: Python is similarly behind rust as this PR addresses, source here: https://github.com/livekit/rust-sdks/blob/c4b5f7f53ad5aeb92e7539c0aeed42f93f449024/livekit/src/room/mod.rs#L402

PR #171, opened 2026-06-14
Fix FfiClient listener use-after-free during Room teardown

alan-george-lk:
Fixes a use-after-free crash when a Room (or any listener owner) is destroyed while the FFI thread is still dispatching an event into its FfiClient listener. Observed both in flaky integration tests and user-reported.

- Introduce `ListenerSlot` (additional state around `Listener`) to track in-flight listener callbacks; removeListener() and shutdown() block until active callbacks finish before teardown proceeds
- Set a `removed` flag  before draining so stale pushEvent snapshots do not start new callbacks after unregister
- Replace the `initialized_` bool with a `LifecycleState` enum and compare_exchange_strong for single-flight initialize() / shutdown(), and to drop events / reject new listeners during shutdown
- Added 2 unit tests to verify fixes are resiliant

### DIFF
--- a/src/ffi_client.cpp
+++ b/src/ffi_client.cpp
@@ -151,10 +151,7 @@ FfiClient::~FfiClient() {
   if (lifecycle_state_.load() == LifecycleState::Initialized) {
     // Explicitly use this over spdlog
     // spdlog can throw, and wrapping in try/catch also flags "empty catch" clang-tidy check
-    std::fputs(
-        "LiveKit SDK was not shut down before process exit. "
-        "Call livekit::shutdown().\n",
-        stderr);
+    std::fputs("[livekit] [warning] SDK was not shut down before process exit. Use livekit::shutdown()\n", stderr);
     std::fflush(stderr);
   }
 }
@@ -165,10 +162,12 @@ void FfiClient::shutdown() noexcept {
     // Atomically claim shutdown ownership; only the caller that transitions
     // Initialized -> ShuttingDown may drain callbacks and dispose the FFI.
     LifecycleState expected = LifecycleState::Initialized;
-    // Note: compare_exchange_strong transitions Initialized -> ShuttingDown
     if (!lifecycle_state_.compare_exchange_strong(expected, LifecycleState::ShuttingDown, std::memory_order_acq_rel)) {
+      // If not Initialized, return early to avoid unnecessary cleanup
       return;
     }
+
+    // Initialized, proceed with cleanup
     dispose_ffi = true;
 
     std::vector<std::shared_ptr<ListenerSlot>> listeners_to_drain;
@@ -179,15 +178,18 @@ void FfiClient::shutdown() noexcept {
       for (auto& [id, slot] : listeners_) {
         (void)id;
         if (slot) {
+          // Mark the listener as removed to prevent race conditions
           {
             const std::scoped_lock<std::mutex> slot_guard(slot->mutex);
             slot->removed = true;
           }
+          // Add the listener to the list of listeners to drain
           listeners_to_drain.push_back(std::move(slot));
         }
       }
       listeners_.clear();
 
+      // Add the pending operations to the list of pending operations to cancel
       pending_to_cancel.reserve(pending_by_id_.size());
       for (auto& [async_id, pending] : pending_by_id_) {
         (void)async_id;
@@ -198,13 +200,19 @@ void FfiClient::shutdown() noexcept {
       pending_by_id_.clear();
     }
 
+    // Cancel the pending operations
     for (auto& pending : pending_to_cancel) {
       pending->cancel();
     }
 
     const auto this_thread = std::this_thread::get_id();
+    // Wait for all in-flight listener callbacks to complete
     for (const auto& slot : listeners_to_drain) {
       std::unique_lock<std::mutex> slot_lock(slot->mutex);
+
+      // When shutdown() isn't on a listener thread, self_active is 0 and we wait for active_callbacks == 0. When it's
+      // called from inside a listener (e.g. ShutdownFromListenerDoesNotDeadlock), self_active is 1 and the wait
+      // succeeds immediately with active_callbacks == 1, so we don't wait on our own in-flight callback.
       slot->cv.wait(slot_lock, [&slot, this_thread] {
         const auto thread_it = slot->active_threads.find(this_thread);
         const int self_active = thread_it == slot->active_threads.end() ? 0 : thread_it->second;
@@ -215,12 +223,15 @@ void FfiClient::shutdown() noexcept {
     livekit_ffi_dispose();
     dispose_ffi = false;
     lifecycle_state_.store(LifecycleState::Uninitialized, std::memory_order_release);
-  } catch (...) {
+  } catch (const std::exception& e) {
     if (dispose_ffi) {
       livekit_ffi_dispose();
       lifecycle_state_.store(LifecycleState::Uninitialized, std::memory_order_release);
     }
-    (void)std::fputs("LiveKit SDK shutdown failed during local cleanup.\n", stderr);
+    // Explicitly use this over spdlog (method is noexcept)
+    (void)std::fputs("[livekit] [error] SDK shutdown failed during local cleanup: ", stderr);
+    (void)std::fputs(e.what(), stderr);
+    (void)std::fputs("\n", stderr);
     (void)std::fflush(stderr);
   }
 }
--- a/src/ffi_client.cpp
+++ b/src/ffi_client.cpp
@@ -18,6 +18,7 @@
 
 #include <cassert>
 #include <csignal>
+#include <cstdio>
 
 #include "data_track.pb.h"
 #include "e2ee.pb.h"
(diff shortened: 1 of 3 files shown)


src/ffi_client.cpp
```diff
@@ -146,44 +147,145 @@ FfiClient& FfiClient::instance() noexcept {
...
-// due to the assert being pre-processed out
-// NOLINTNEXTLINE(modernize-use-equals-default)
 FfiClient::~FfiClient() {
-  assert(!initialized_.load() &&
-         "LiveKit SDK was not shut down before process exit. "
-         "Call livekit::shutdown().");
+  if (lifecycle_state_.load() == LifecycleState::Initialized) {
+    // Explicitly use this over spdlog
+    // spdlog can throw, and wrapping in try/catch also flags "empty catch" clang-tidy check
+    std::fputs("[livekit] [warning] SDK was not shut down before process exit. Use livekit::shutdown()\n", stderr);
```
stephen-derosa, line 154: why use std::fputs to buffer then flush vs std::cerr which outputs immediately?

src/ffi_client.cpp
```diff
@@ -146,44 +147,145 @@ FfiClient& FfiClient::instance() noexcept {
...
 }
 
 void FfiClient::shutdown() noexcept {
-  if (!isInitialized()) {
-    return;
+  // Don't use string to avoid exceptions
+  // (Also cleaner with exception.what() and printing)
+  const char* shutdown_error = nullptr;
+  try {
+    // Atomically claim shutdown ownership; only the caller that transitions
```
stephen-derosa: I find semi colons in code comments to be pretty confusing, consider replacing with
`    // Atomically claim shutdown ownership so only the caller that transitions` or something similar

src/ffi_client.cpp
```diff
@@ -146,44 +147,145 @@ FfiClient& FfiClient::instance() noexcept {
...
+    // Initialized -> ShuttingDown may drain callbacks and dispose the FFI.
+    LifecycleState expected = LifecycleState::Initialized;
+    if (!lifecycle_state_.compare_exchange_strong(expected, LifecycleState::ShuttingDown, std::memory_order_acq_rel)) {
+      // If not Initialized, return early to avoid unnecessary cleanup
```
stephen-derosa, line 168: should we log here?

src/ffi_client.cpp
```diff
@@ -146,44 +147,145 @@ FfiClient& FfiClient::instance() noexcept {
...
 }
 
 void FfiClient::removeListener(ListenerId id) {
-  const std::scoped_lock<std::mutex> guard(lock_);
-  listeners_.erase(id);
+  std::shared_ptr<ListenerSlot> slot;
+  {
+    const std::scoped_lock<std::mutex> guard(lock_);
+    auto it = listeners_.find(id);
+    if (it == listeners_.end()) {
```
stephen-derosa, line 277: is it worth logging here? will this ever happen?

src/ffi_client.cpp
```diff
@@ -146,44 +147,145 @@ FfiClient& FfiClient::instance() noexcept {
...
+      return;
+    }
+    slot = std::move(it->second);
+    listeners_.erase(it);
+  }
+
+  const auto this_thread = std::this_thread::get_id();
+  std::unique_lock<std::mutex> slot_lock(slot->mutex);
+  slot->removed = true;
```
stephen-derosa: shouldnt `slot->removed = true` come after the `slot->cv.wait()` call?

src/ffi_client.cpp
```diff
@@ -246,8 +351,39 @@ void FfiClient::pushEvent(const proto::FfiEvent& event) const {
...
+      const auto thread_it = slot->active_threads.find(this_thread);
+      if (thread_it != slot->active_threads.end()) {
+        --thread_it->second;
+        if (thread_it->second == 0) {
+          slot->active_threads.erase(thread_it);
+        }
+      }
+      --slot->active_callbacks;
+    }
+    slot->cv.notify_all();
```
stephen-derosa, line 390: consider commenting `notify the CV in case it is waiting to be released` or something

src/tests/unit/test_ffi_client.cpp
```diff
@@ -38,6 +42,76 @@ void handleSignal(int signal) {
   }
 }
 
+void emitLogEvent() {
```
stephen-derosa: ah is the std::fputs for testing? :mind-blown:

src/tests/unit/test_ffi_client.cpp
```diff
@@ -144,15 +218,174 @@ TEST_F(FfiClientTest, RemoveListenerIsIdempotent) {
...
+  std::atomic<bool> callback_completed{false};
+
+  const auto id = FfiClient::instance().addListener([&](const proto::FfiEvent&) {
+    callback_entered.set_value();
+    release_callback_future.wait();
+    callback_completed.store(true);
+  });
+
+  std::thread callback_thread([] { emitLogEvent(); });
+  ASSERT_EQ(callback_entered_future.wait_for(std::chrono::seconds(5)), std::future_status::ready);
```
stephen-derosa: nit for all these tests with `wait_for()` -- add a comment of why we are waiting that long/short

src/ffi_client.cpp
```diff
@@ -146,44 +147,145 @@ FfiClient& FfiClient::instance() noexcept {
...
+    (void)std::fputs(shutdown_error, stderr);
+    (void)std::fputs("\n", stderr);
+    (void)std::fflush(stderr);
+  }
 }
 
 bool FfiClient::initialize(bool capture_logs) {
-  if (isInitialized()) {
+  LifecycleState expected = LifecycleState::Uninitialized;
+  if (!lifecycle_state_.compare_exchange_strong(expected, LifecycleState::Initializing, std::memory_order_acq_rel)) {
```
stephen-derosa, line 243: a comment on why `compare_exchange_strong` here would be nice

stephen-derosa, line 243: a comment on why `compare_exchange_strong` here would be nice

stephen-derosa (approved):
nice this looks good -- thank you for driving this down!

testing is king ofc, and the unit tests seem pretty wholesome! Just given that we know this should addresses instability and that a customer will immediately use this, I would suggest a more thorough body of the PR and  adding something about a `Listener state machine` (or something similar) to the commit.

src/ffi_client.cpp
alan-george-lk: In a destructor clang-tidy flags any lines that throw: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/exception-escape.html

Although rare both `spdlog` and `std::cerr` can throw, hence `std::fputs` and flush. I don't love the yet third logging option here, but alas.

I also used it in `shutdown()` for this same reason as it was marked `noexpect` already prior to this PR

src/ffi_client.cpp
alan-george-lk: Updated

src/ffi_client.cpp
alan-george-lk: Good call, done

src/ffi_client.cpp
alan-george-lk: I think given the mutex position this shouldn't happen, open to adding a log but I think this is less critical than other error state logs

src/ffi_client.cpp
alan-george-lk: Fair through process but actually not. It's a bit confusing:

- This flag is set so that `pushEvent()` can't push any more callback invocations for this listener (local to ffi_client) version after it's marked as removed, which is copied prior to the `slot->mutex` lock (within pushEvent itself) which would syncronize this flag
- The wait drains in-flight callbacks before continuing (local to ffi_client listener)
- If it was the other way around, `pushEvent()` could be hit and the listener copied again even though it was marked as removal

This (slot mutex + flag approach) is kind of an ugly side effect of the "copy listeners before invoking them strategy" to avoid deadlock if the user called SDK operations from within a callback

src/ffi_client.cpp
alan-george-lk: Added:

```
// Notify in case this listener was marked for removal during the callback (will be waiting on this)
```

src/tests/unit/test_ffi_client.cpp
alan-george-lk: Ha, not quite! This is a simple helper to invoke (but mainly spam) log events to verify in flight callbacks get drained/no crash or deadlock when shutdown starts. See here for `fputs` reasoning: https://github.com/livekit/client-sdk-cpp/pull/171#discussion_r3415888899

Even though FFI log forwarding isn't setup in this branch, it still gets through the `pushEvent` pipeline queue and just gets ignored since it's not a supported event type: https://github.com/livekit/client-sdk-cpp/blob/main/src/ffi_client.cpp#L267

src/tests/unit/test_ffi_client.cpp
alan-george-lk: Moved all the 5 seconds to a single const variable and put the comment up with it

src/ffi_client.cpp
alan-george-lk: Generally it's just good practice when transitioning an atomic to do this vs. a load() and store() separately, which are two atomic operations where this is one. Happy to still add a comment if you prefer

alan-george-lk: See https://github.com/livekit/client-sdk-cpp/pull/171#discussion_r3416168602

src/ffi_client.cpp
stephen-derosa: nice thanks for the info!

src/ffi_client.cpp
stephen-derosa: sgtm

src/ffi_client.cpp
stephen-derosa: not required, up to you

PR #177, opened 2026-06-18
Add token source API for fetching LiveKit tokens

alan-george-lk:
## Summary

- Adds new public API `token_source.h`, allowing users to now mint tokens using the following paths:
  -  Literal: user-provided token (basically the legacy `livekit-cli` minted path)
  - Sandbox: enable a sandbox token server via LiveKit cloud, and allow users insecurely mint tokens (development only)
  - Endpoint: production-level server support
  - Custom: additional user token minting support
- Additional public API changes include: 
  - `room.h` -> new ways to connect with tokens
  - `room_delegate.h` -> new callback with the new event
  - `room_event_types.h` -> new information-only event type when tokens are refreshed by the server (see below)

### Server Token Refresh

LiveKit server token refresh existed before this feature, but this branch now adds support for the FFI event `onTokenRefreshed`.

## New Dependencies

- [libcurl4-openssl](https://packages.debian.org/sid/libcurl4-openssl-dev) - used for HTTP token endpoint support
- [nlohmann/json](https://github.com/nlohmann/json) - JSON serialization/deserialization support for all configurable token source types

## Testing

Validated through a combination of unit, integration, and standalone binary tester program against a LiveKit Cloud agent with a live token server.

### Token Server Action

Integration tests now test against a basic [token server implementation](https://github.com/livekit/token-server-action) run in GitHub actions live (skipped locally).

### Sandbox Token

Verified locally by connecting to a LiveKit cloud instance:

```
auto sandbox_token = livekit::SandboxTokenSource::create("<omitted but valid>");
```

And results printed below:

```
./build-release/bin/token_source_tester
...
Server URL: <omitted but valid>
Participant Token: <omitted but valid>
Participant Name: efficient-sensor
Connected to room
Local participant: cpp-test-a
```


include/livekit/room.h
```diff
@@ -123,6 +124,12 @@ struct RoomOptions {
   std::optional<std::chrono::milliseconds> connect_timeout;
 };
 
+/// Async callback that supplies an access token for room connection.
+///
+/// Invoked on the application thread during @ref Room::connect before the FFI
+/// connect request is sent. The returned future must resolve to a non-empty JWT.
+using TokenSource = std::function<std::future<std::string>()>;
```
xianshijing-lk: + @MaxHeimbrock  

I don't really understand this TokenSource impl, is it supposed to be implemented by our developers ? or we can use some implementation in Rust ?

xianshijing-lk (commented):
might need help to understand the TokenSource a bit more

include/livekit/room.h
alan-george-lk: This was a bad first pass -- I reviewed the code last night and felt the same. Working on a proper implementation and will request your review when _actually_ ready :)

MaxHeimbrock:
I am using the tokensource sandbox and it is working fine for me 👍

docs/token-lifecycle.md
```diff
@@ -0,0 +1,109 @@
+# Token lifecycle
+
+Succinct reference for how join credentials and in-session refresh interact in
```
xianshijing-lk: @1egoman , could you please help review this token-lifecycle.md ?

include/livekit/room.h
```diff
@@ -28,6 +28,7 @@
 #include "livekit/room_event_types.h"
 #include "livekit/stats.h"
 #include "livekit/subscription_thread_dispatcher.h"
+#include "livekit/token_source.h"
```
xianshijing-lk: nit, would you consider forward declare those types, and move the #include "livekit/token_source.h" to the cpp ? 

One benefit for it is that it could speed up the build.

include/livekit/token_source.h
```diff
@@ -0,0 +1,310 @@
...
+
+/// @brief Error returned when token fetching fails.
+struct TokenSourceError {
+  std::string message;
+};
+
+/// @brief Base interface for token sources that provide full credentials directly.
+class LIVEKIT_API TokenSourceFixed {
+public:
+  virtual ~TokenSourceFixed();
```
xianshijing-lk: nit, just do "virtual ~TokenSourceFixed() = default" here and remove the cpp line

The same for  virtual ~TokenSourceConfigurable() = default;

include/livekit/token_source.h
```diff
@@ -0,0 +1,310 @@
...
+  /// Additional request headers.
+  std::map<std::string, std::string> headers;
+
+  /// Request timeout (default 30 seconds).
+  std::chrono::milliseconds timeout = std::chrono::seconds(30);
+};
+
+/// @brief Error returned when token fetching fails.
+struct TokenSourceError {
+  std::string message;
```
xianshijing-lk, line 150: nit, do we also want the http status_code ?

src/room.cpp
```diff
@@ -91,6 +91,49 @@ void Room::setDelegate(RoomDelegate* delegate) {
...
+
+  if (!details) {
+    LK_LOG_ERROR("Room::connect failed: token source error: {}", details.error().message);
+    return false;
+  }
+
+  return connect(details.value().server_url, details.value().participant_token, options);
+}
+
+bool Room::connect(TokenSourceConfigurable& token_source, const TokenRequestOptions& request_options,
```
xianshijing-lk: consider a helper function like 
```
namespace {

template <typename FetchFn>
bool connectWithTokenSource(Room& room, const RoomOptions& options, FetchFn&& fetch) {
  Result<TokenSourceResponse, TokenSourceError> details =
      Result<TokenSourceResponse, TokenSourceError>::failure(TokenSourceError{"token source not invoked"});

  try {
    details = fetch().get();
  } catch (const std::exception& e) {
    LK_LOG_ERROR("Room::connect failed: token source threw: {}", e.what());
    return false;
  } catch (...) {
    LK_LOG_ERROR("Room::connect failed: token source threw unknown exception");
    return false;
  }

  if (!details) {
    LK_LOG_ERROR("Room::connect failed: token source error: {}", details.error().message);
    return false;
  }

  const auto& value = details.value();
  return room.connect(value.server_url, value.participant_token, options);
}

} // namespace
```

Then these two function can become:
```
bool Room::connect(TokenSourceFixed& token_source, const RoomOptions& options) {
  return connectWithTokenSource(*this, options, [&] {
    return token_source.fetch();
  });
}

bool Room::connect(TokenSourceConfigurable& token_source,
                   const TokenRequestOptions& request_options,
                   const RoomOptions& options) {
  return connectWithTokenSource(*this, options, [&] {
    return token_source.fetch(request_options, false);
  });
}
```

src/token_source.cpp
```diff
@@ -0,0 +1,264 @@
...
+    std::function<std::future<Result<TokenSourceResponse, TokenSourceError>>(const TokenRequestOptions&)> provider) {
+  return std::unique_ptr<CustomTokenSource>(new CustomTokenSource(std::move(provider)));
+}
+
+CustomTokenSource::CustomTokenSource(
+    std::function<std::future<Result<TokenSourceResponse, TokenSourceError>>(const TokenRequestOptions&)> provider)
+    : provider_(std::move(provider)) {}
+
+std::future<Result<TokenSourceResponse, TokenSourceError>> CustomTokenSource::fetch(const TokenRequestOptions& options,
+                                                                                    bool /*force_refresh*/) {
```
xianshijing-lk: any idea why CustomTokenSource will skip the force_refresh ?

src/token_source.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+  return runAsyncTokenSource("token source cache fetch failed", [this, options_snapshot, force_refresh]() {
+    const std::scoped_lock<std::mutex> lock(mutex_);
+    if (!force_refresh && cached_details_.has_value() && cached_options_.has_value() &&
+        tokenRequestOptionsEqual(*cached_options_, *options_snapshot) &&
+        isParticipantTokenValid(cached_details_->participant_token)) {
+      return TokenSourceResult::success(*cached_details_);
+    }
+
+    auto result = inner_->fetch(*options_snapshot, force_refresh).get();
```
xianshijing-lk: any reason why we want to run the inner_->fetch() under the lock ? 

Will it be safer to do ?
```
{
  std::scoped_lock lock(mutex_);
  if (!force_refresh && cached_details_ && cached_options_ &&
      tokenRequestOptionsEqual(*cached_options_, *options_snapshot) &&
      isParticipantTokenValid(cached_details_->participant_token)) {
    return TokenSourceResult::success(*cached_details_);
  }
}

auto result = inner_->fetch(*options_snapshot, force_refresh).get();
```

src/token_source.cpp
```diff
@@ -0,0 +1,264 @@
...
+}
+
+std::unique_ptr<SandboxTokenSource> SandboxTokenSourceTestAccess::create(const std::string& sandbox_id,
+                                                                         TokenEndpointOptions options,
+                                                                         const std::string& base_url,
+                                                                         TokenSourceHttpTransport transport) {
+  auto source = std::unique_ptr<SandboxTokenSource>(new SandboxTokenSource(sandbox_id, options, base_url));
+  auto resolved = resolveSandboxEndpoint(sandbox_id, std::move(options), base_url);
+  source->endpoint_ =
+      EndpointTokenSourceTestAccess::create(std::move(resolved.url), std::move(resolved.options), std::move(transport));
```
xianshijing-lk, line 219: isn't the endpoint_ created in line 210 already ? any reason why it needs to be re-created ?

src/token_source_http.cpp
```diff
@@ -0,0 +1,242 @@
...
+  components.dwUrlPathLength = static_cast<DWORD>(-1);
+  components.dwExtraInfoLength = static_cast<DWORD>(-1);
+
+  const std::wstring wide_url = toWide(url);
+  if (!WinHttpCrackUrl(wide_url.c_str(), 0, 0, &components)) {
+    return Result<std::string, std::string>::failure("failed to parse token server URL");
+  }
+
+  const std::wstring host(components.lpszHostName, components.dwHostNameLength);
+  const std::wstring path(components.lpszUrlPath, components.dwUrlPathLength);
```
xianshijing-lk: question, does path need to include lpszExtraInfo ? like when lpszExtraInfoLength > 0 ?

src/token_source_http.cpp
```diff
@@ -0,0 +1,242 @@
...
+  const std::wstring wide_method = toWide(normalizeHttpMethod(method));
+
+  HINTERNET session = WinHttpOpen(L"LiveKit-CPP/1.0", WINHTTP_ACCESS_TYPE_DEFAULT_PROXY, WINHTTP_NO_PROXY_NAME,
+                                  WINHTTP_NO_PROXY_BYPASS, 0);
+  if (session == nullptr) {
+    return Result<std::string, std::string>::failure("WinHttpOpen failed");
+  }
+
+  const int timeout_ms = static_cast<int>(timeout.count());
+  WinHttpSetTimeouts(session, timeout_ms, timeout_ms, timeout_ms, timeout_ms);
```
xianshijing-lk, line 113: what will happen if timeout_ms is 0 or not valid ?

xianshijing-lk (approved):
lgtm with some comments, please address them.

Great work!!!

docs/token-lifecycle.md
alan-george-lk: I actually generated this for him while I had some active LLM context around this topic, I removed it as I don't want this SDK to be an authoritative documentation source compared to docs.livekit.io.

I do think we should have some better docs around token source and lifetimes, while going through this process it felt a little disjointed and agents/frontend-focused and not just a general document for token source terms, functions, lifecycles, etc for client SDKs.

include/livekit/token_source.h
alan-george-lk: Reasonable idea, but I would advocate not doing this. This error applies to all token source types (including literal and custom, which have no external comms), so adding specific details to any one implementation wouldn't be clean in my view.

If we get a user request for HTTP details for some reason, we could extend or more generalize error specifics into this struct.

include/livekit/token_source.h
```diff
@@ -0,0 +1,310 @@
...
+  /// Often JSON. Applies to the remote agent worker, not the local participant
+  /// (use @c participant_metadata for that). Ignored unless @c agent_name is set
+  /// or another agent field triggers @c room_config serialization.
+  std::optional<std::string> agent_metadata;
+
+  /// LiveKit Cloud deployment to target for agent dispatch.
+  ///
+  /// Optional. When omitted or empty, the production deployment is used.
+  /// Only relevant when dispatching a named agent on LiveKit Cloud.
+  std::optional<std::string> agent_deployment;
```
1egoman, line 124: nitpick: in other sdks (for example, [web](/Users/ryan/w/livekit/client-sdk-js/.changeset/slick-geese-happen.md)) this was just called `deployment`. I like this `agentDeployment` name better though and had brought it up during the review process when it was added recently but I think it got missed. It might be worth changing this for consistency, idk.

include/livekit/token_source.h
```diff
@@ -0,0 +1,310 @@
...
+  /// Each @ref fetch call returns the same credentials.
+  ///
+  /// @param server_url WebSocket URL of the LiveKit server.
+  /// @param participant_token JWT access token for the participant.
+  static std::unique_ptr<LiteralTokenSource> fromValue(std::string server_url, std::string participant_token);
+
+  /// @brief Create a token source from an async provider that returns full credentials.
+  ///
+  /// Use this overload when credentials are produced outside the SDK but fetched
+  /// lazily (for example, from your own cache or secure storage).
```
1egoman, line 196: nitpick: It might be worth re-emphasizing in the docs here (assuming this works the same as the web and other implementations) that this doesn't support parameters and that if you want a configurable version of this which can re-fetch multiple times on demand to use `CustomTokenSource`

src/token_source.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+using TokenSourceResult = Result<TokenSourceResponse, TokenSourceError>;
+using TokenSourceFuture = std::future<TokenSourceResult>;
+
+bool tokenRequestOptionsEqual(const TokenRequestOptions& a, const TokenRequestOptions& b) {
+  return a.room_name == b.room_name && a.participant_name == b.participant_name &&
+         a.participant_identity == b.participant_identity && a.participant_metadata == b.participant_metadata &&
+         a.participant_attributes == b.participant_attributes && a.agent_name == b.agent_name &&
+         a.agent_metadata == b.agent_metadata && a.agent_deployment == b.agent_deployment;
+}
```
1egoman, line 38: nitpick: I'm not sure if there's a way to handle this in c++, but ideally if possible it would be great if the compiler could somehow throw an error if a field in `a` or `b` were missed in this check.

Also, do you have to worry about deep equality here within maps like `participant_attributes`? Or is `==` on the fields directly good enough?

include/livekit/token_source.h
```diff
@@ -0,0 +1,310 @@
...
+
+  friend struct SandboxTokenSourceTestAccess;
+};
+
+/// @brief Decorator that adds JWT-aware caching to another configurable token source.
+///
+/// Wrap @ref CustomTokenSource, @ref EndpointTokenSource, or
+/// @ref SandboxTokenSource to reduce token fetch calls while still refreshing
+/// when tokens expire or when @p force_refresh is requested.
+class LIVEKIT_API CachingTokenSource final : public TokenSourceConfigurable {
```
1egoman, line 318: thought: You probably have seen it / maybe even patterned it off of this, but your implementation here is fairly close to the swift version: https://github.com/livekit/client-sdk-swift/blob/5d13e9ca7b366a8b47bd08c95a34cb0068149287/Sources/LiveKit/Token/CachingTokenSource.swift#L23.

Also just want to confirm - looking at the cpp I think this is just memoization with the last value correct? There's no other way to configure this right? I think that fine.

1egoman (commented):
Generally looks good to me, and I think matches the interfaces on other sdks pretty well!

src/room.cpp
1egoman: FWIW, we opted not to do this in other sdks, and determined this was the session api's job. Relevant pull request: https://github.com/livekit/client-sdk-js/pull/1677

src/token_source.cpp
alan-george-lk: This comment actually led to an API design change, see here: https://github.com/livekit/client-sdk-cpp/commit/e39e55ec01580e3bbcd126fe73c4e8fd95f1d7b0

Having such an argument for a custom source would require passing that argument along to the custom implementation, which felt weird as the custom token source was just a user function that could decide anything it needed to. I have instead dropped `force_refresh` entirely, and opted for `caching_token_source.invalidate()` which matches how Swift/Android do it.

In short, this original design around forcing refresh was modeled after JS with the `force_refresh`. But JS actually goes for an inheritance-based approach to cacheing token sources, vs. here which is using a decorator (Swift/Android do it this way), so forcing refresh was consistent across all sub-classes and didn't have an awkward drop like this did.

It was basically a divergence in both approaches (force + decorator vs. force + inheritance), now it's more aligned to Swift/Android.

src/room.cpp
alan-george-lk: Added!

include/livekit/token_source.h
alan-george-lk: Added

include/livekit/room.h
alan-george-lk: Added

src/token_source.cpp
alan-george-lk: Holding the mutex over the fetch is how JS and Swift do it, I'm open to your change but it would be a divergence

src/token_source.cpp
alan-george-lk: Great catch, this slipped through. Now have a private constructor that accepts an already made endpoint so both the public and test access versions can avoid recreation

src/token_source_http.cpp
alan-george-lk: Addressed

src/token_source_http.cpp
alan-george-lk: Good call, <= 0 now goes to the default of 30 seconds

include/livekit/token_source.h
alan-george-lk: Android and Swift (which this largely models) both use `agentDeployment`. So I think it's a 50/50 consistency roll of the dice 🤔

1egoman: Ugh I hadn't realized it was so fragmented 😧

src/token_source.cpp
alan-george-lk: For the second part, `==` is an "operator overload", which is a C++-ism that basically allows you to define actual functions for operators like +, -, *, ==, etc., so for maps it is in fact a deep comparison: https://en.cppreference.com/cpp/container/map/operator_cmp (good callout though!)

For the first one, there are some options, I am looking into that

include/livekit/token_source.h
alan-george-lk: Added!

src/token_source.cpp
alan-george-lk: @1egoman For the first part, it's a good callout, but until C++20 which now has default bool operators (compiler-generated field quality checks like this does manually), what you see here is basically the way to do it unfortunately. There's some C++17 (what we're on) features to slightly improve the robustness, but it adds a lot of code for little benefit. 

My vote would be to leave this as-is as it's unlikely to change often

include/livekit/token_source.h
alan-george-lk: Yup and yup to your two questions!

PR #178, opened 2026-06-18
Minimize docker build

alan-george-lk:
- Limits Docker operations to only the `docker-images.yml` workflow (not in `builds.yml` anymore)
- Docker only runs when Docker-related stuff changes
- Nightly CI will validate docker builds more regularly (downstream PR)


stephen-derosa (approved):
🔥

PR #179, opened 2026-06-22
Enable integration tests on Windows

alan-george-lk:
- Enables integration tests on Windows
- Adds additional guard in PlatformAudio test to skip on Windows


stephen-derosa (approved):
nice!

PR #195, opened 2026-07-06
subscription_thread_dispatcher: better callback/thread lifecycle

### LINKED ISSUE #235: Calling setOnVideo/AudioFrameCallback outside of the onTrackSubscribed handler never starts a frame thread handler
### LiveKit C++ SDK Version

v1.2.0

### Platform

Linux x86_64

### OS Version

OpenSUSE 15.6 LEAP

### Compiler & Version

gcc 15.2

### CMake Version

3.28.3

### How are you using the SDK?

Vendoring the SDK and building from source as a CMake subdirectory

### Issue Description

Calling `setOnAudio/VideoFrameCallback` correctly adds the frame handler to the relevant `video/audio_callbacks`, but doing so outside of the `onTrackSubscribed` delegate callback never starts a reader thread, resulting in the corresponding frame callback never being called. 

This happens because `SubscriberThreadDispatcher::handleTrackSubscribed` only gets called synchronously right after the delegate callback from the Room event handler @ `room.cpp:703`, which no-ops if the frame callback key has not been added to the callback map yet.

A potential fix could be to call `SubscriberThreadDispatcher::handleTrackSubscribed` if `setOnAudio/VideoFrameCallback` is called when the mappings are empty.

### Steps to Reproduce

1. Handle the `onTrackSubscribed` room delegate event
2. Call `Room::setOnVideoFrameCallback` from the context of another thread (ex: in a Qt GUI thread) that can run after the event handler returns
3. Notice no audio/video frame callbacks being called

stephen-derosa:
# Overview

Fixes #235 and hardens reader-thread lifecycle in `SubscriptionThreadDispatcher`. Frame callbacks now start a reader regardless of when they are registered, replacement actually replaces, and driving the dispatcher from inside a frame callback no longer self-joins. **No public API changes** — `setOn*FrameCallback` now does what it always documented.

## The bugs

- **Late registration never started a reader (#235).** `handleTrackSubscribed` ran once, right after `RoomDelegate::onTrackSubscribed`, and no-oped without a registered callback. Registering afterwards (e.g. from a GUI thread) stored the callback and nothing else happened.
- **Replacement was a silent no-op.** Reader threads capture the callback **by value**. Re-registering only overwrote the dispatcher's copy; the running thread kept invoking the old callback.
- **Self-joins terminated the process.** `Room::disconnect()` or `removeOnDataFrameCallback` called from inside a data frame callback joined the calling thread: `std::system_error`, then a still-joinable `std::thread` destroyed during unwinding.

## Changes

- **Retained subscriptions.** The dispatcher keeps the subscribed track per `(participant, track_name)` until unsubscribed, so `setOn*` starts a reader immediately when the track is already subscribed and defers otherwise. `Room` also resolves the current publication, which covers registration from *inside* `onTrackSubscribed`.
- **Replace in place, join before start.** Every stop path (`setOn*`, `clearOn*`, resubscribe with a new SID, unsubscribe) extracts the reader under the lock, marks the key *draining*, joins outside the lock, then restarts. No caller can start a reader for a draining key, so old and new callbacks never run concurrently and concurrent setters cannot pile up untracked threads.
- **Duplicate subscription events.** A repeated `track_subscribed` for the SID a reader already serves is a no-op; a new SID (republish) stops the old reader first.
- **Detach on self-join.** No reader lambda captures `this`; data readers report exit through their own shared state. All readers are therefore safe to detach, so a lifecycle call from inside the reader's own callback logs a warning, detaches, and still takes effect.
- **Data track teardown.** In-flight `RemoteDataTrack::subscribe()` calls are marked cancelled so unpublish/republish cannot leave a stale reader or duplicate FFI subscription behind.
- **`SubscriptionThreadDispatcher` deprecated** for direct use (`LIVEKIT_DEPRECATED`); it stays exported to avoid a major bump.

## Behavioral notes

- **`setOn*`/`clearOn*` block** until any in-flight invocation of the previous callback returns. A callback that never returns blocks registration indefinitely.
- **Re-entrant calls are discouraged but handled.** From inside a frame callback for the same key the SDK warns and detaches that reader. The in-flight invocation of the previous callback finishes only after the call returns, so the no-overlap guarantee does not hold for that one invocation.

## Public API

**No signature changes.** `include/livekit/room.h` changes are documentation only. Callback type aliases moved to `include/livekit/frame_callbacks.h` (still reachable through `room.h`).

**Test-only:** `RemoteDataTrack::testFfiHandleId()` removed in favor of `RemoteDataTrackTestAccess`; `RoomTestAccess` exposes reader/drain counts and retained-subscription state; `RemoteDataTrack`'s private constructor is `LIVEKIT_INTERNAL_API` so tests can link it.

## Testing

- **Unit** (`SubscriptionThreadDispatcherTest`, 91): #235 in both orderings for audio/video/video-event; replacement semantics including the shared video slot; SID dedup and republish; drain marks set and cleared by every lifecycle path, with starts deferred while draining; self-join handling for `setOn*`, `clearOn*`, `removeOnDataFrameCallback`, and `stopAll` from both media and data reader threads; data reader cancellation and finished-state handling.
- **Integration** (`FrameCallbackServerTest`, 5): the exact #235 scenario for video, video-event, and audio; clear-and-re-register without a new event; registration from inside `onTrackSubscribed` starting exactly one reader.
- **Integration** (`FrameCallbackReplacementTest`, 14): delivery switches on replacement; setter blocks for a slow callback without overlapping it; sequential and concurrent churn leave exactly one reader; deferred start and republish; re-entrant `setOn*`/`clearOn*`/data removal take effect; `Room::disconnect()` from inside a data callback.
- Full unit suite (412) and the data-track, room-event, and platform-audio integration suites (37) pass locally against a LiveKit server. clang-format and clang-tidy are clean on the changed files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,13 @@ The following features are deprecated and will be removed in the next major rele
 
 - `PacketTrailerFeatures` is deprecated. Use `FrameMetadataFeatures` via
   `TrackPublishOptions::frame_metadata_features` instead.
+- `Room::setOnAudioFrameCallback`, `Room::setOnVideoFrameCallback`, and
+  `Room::setOnVideoFrameEventCallback` are deprecated. Use the `[[nodiscard]]`
+  variants `trySetOnAudioFrameCallback`, `trySetOnVideoFrameCallback`, and
+  `trySetOnVideoFrameEventCallback` instead, which return `false` when a reader
+  is already active for the key (instead of silently replacing a running callback).
+  To replace an active callback, call `clearOn*FrameCallback` first. (The same rename
+  applies to the corresponding `SubscriptionThreadDispatcher` methods.)
 
 ### `v1.0.0`
 
--- a/include/livekit/remote_data_track.h
+++ b/include/livekit/remote_data_track.h
@@ -92,11 +92,6 @@ public:
   /// @param options Pipeline options to apply to this remote data track.
   LIVEKIT_API void setPipelineOptions(const DataTrackPipelineOptions& options);
 
-#ifdef LIVEKIT_TEST_ACCESS
-  /// Test-only accessor for exercising lower-level FFI subscription paths.
-  uintptr_t testFfiHandleId() const noexcept { return ffiHandleId(); }
-#endif
-
   /// Subscribe to this remote data track.
   ///
   /// Returns a DataTrackStream that delivers frames via blocking
@@ -106,8 +101,9 @@ public:
 
 private:
   friend class Room;
+  friend struct RemoteDataTrackTestAccess;
 
-  explicit RemoteDataTrack(const proto::OwnedRemoteDataTrack& owned);
+  LIVEKIT_INTERNAL_API explicit RemoteDataTrack(const proto::OwnedRemoteDataTrack& owned);
 
   uintptr_t ffiHandleId() const noexcept { return handle_.get(); }
   /// RAII wrapper for the Rust-owned FFI resource.
--- a/include/livekit/room.h
+++ b/include/livekit/room.h
@@ -303,16 +303,92 @@ public:
   // Frame callbacks
   // ---------------------------------------------------------------
 
-  /// @brief Sets the audio frame callback via SubscriptionThreadDispatcher.
+  /// Register an audio frame callback for a remote subscription.
+  ///
+  /// The callback is keyed by @p participant_identity and @p track_name. If the
+  /// matching remote audio track is already subscribed, a reader is started
+  /// immediately; otherwise the reader starts when the track is subscribed.
+  ///
+  /// To replace a callback whose reader is already running, call
+  /// @ref clearOnAudioFrameCallback first, then register again:
+  /// @code
+  /// room.clearOnAudioFrameCallback(identity, track_name);
+  /// if (!room.trySetOnAudioFrameCallback(identity, track_name, new_handler)) {
+  ///   // registration was rejected (a reader is still active)
+  /// }
+  /// @endcode
+  ///
+  /// @param participant_identity Identity of the remote participant.
+  /// @param track_name           Track name to match.
+  /// @param callback             Function invoked for each decoded audio frame.
+  /// @param opts                 Options used when creating the backing
+  ///                             @ref AudioStream.
+  /// @return @c true if the callback was registered; @c false if a reader is
+  ///         already active for the key (call @ref clearOnAudioFrameCallback
+  ///         first) or the room has no dispatcher.
+  [[nodiscard]] bool trySetOnAudioFrameCallback(const std::string& participant_identity, const std::string& track_name,
+                                                AudioFrameCallback callback, const AudioStream::Options& opts = {});
+
+  /// Register a video frame callback for a remote subscription.
+  ///
+  /// The callback is keyed by @p participant_identity and @p track_name. If the
+  /// matching remote video track is already subscribed, a reader is started
+  /// immediately; otherwise the reader starts when the track is subscribed.
+  ///
+  /// To replace a callback whose reader is already running, call
+  /// @ref clearOnVideoFrameCallback first, then register again.
+  ///
+  /// @param participant_identity Identity of the remote participant.
+  /// @param track_name           Track name to match.
+  /// @param callback             Function invoked for each decoded video frame.
+  /// @param opts                 Options used when creating the backing
+  ///                             @ref VideoStream.
+  /// @return @c true if the callback was registered; @c false if a reader is
+  ///         already active for the key (call @ref clearOnVideoFrameCallback
+  ///         first) or the room has no dispatcher.
+  [[nodiscard]] bool trySetOnVideoFrameCallback(const std::string& participant_identity, const std::string& track_name,
+                                                VideoFrameCallback callback, const VideoStream::Options& opts = {});
+
+  /// Register a rich video frame event callback for a remote subscription.
+  ///
+  /// The callback is keyed by @p participant_identity and @p track_name. If the
+  /// matching remote video track is already subscribed, a reader is started
+  /// immediately; otherwise the reader starts when the track is subscribed.
+  ///
+  /// To replace a callback whose reader is already running, call
+  /// @ref clearOnVideoFrameCallback first, then register again.
+  ///
+  /// @param participant_identity Identity of the remote participant.
+  /// @param track_name           Track name to match.
+  /// @param callback             Function invoked for each decoded video frame
+  ///                             event, including optional metadata.
+  /// @param opts                 Options used when creating the backing
+  ///                             @ref VideoStream.
+  /// @return @c true if the callback was registered; @c false if a reader is
+  ///         already active for the key (call @ref clearOnVideoFrameCallback
+  ///         first) or the room has no dispatcher.
+  [[nodiscard]] bool trySetOnVideoFrameEventCallback(const std::string& participant_identity,
+                                                     const std::string& track_name, VideoFrameEventCallback callback,
+                                                     const VideoStream::Options& opts = {});
+
+  /// @deprecated Use trySetOnAudioFrameCallback() instead.
+  ///
+  /// Forwards to @ref trySetOnAudioFrameCallback and discards the result.
+  [[deprecated("Room::setOnAudioFrameCallback is deprecated; use trySetOnAudioFrameCallback instead")]]
   void setOnAudioFrameCallback(const std::string& participant_identity, const std::string& track_name,
                                AudioFrameCallback callback, const AudioStream::Options& opts = {});
 
-  /// @brief Sets the video frame callback via SubscriptionThreadDispatcher.
+  /// @deprecated Use trySetOnVideoFrameCallback() instead.
+  ///
+  /// Forwards to @ref trySetOnVideoFrameCallback and discards the result.
+  [[deprecated("Room::setOnVideoFrameCallback is deprecated; use trySetOnVideoFrameCallback instead")]]
   void setOnVideoFrameCallback(const std::string& participant_identity, const std::string& track_name,
                                VideoFrameCallback callback, const VideoStream::Options& opts = {});
 
-  /// @brief Sets the video frame event callback via
-  /// SubscriptionThreadDispatcher.
+  /// @deprecated Use trySetOnVideoFrameEventCallback() instead.
+  ///
+  /// Forwards to @ref trySetOnVideoFrameEventCallback and discards the result.
+  [[deprecated("Room::setOnVideoFrameEventCallback is deprecated; use trySetOnVideoFrameEventCallback instead")]]
   void setOnVideoFrameEventCallback(const std::string& participant_identity, const std::string& track_name,
                                     VideoFrameEventCallback callback, const VideoStream::Options& opts = {});
 
@@ -353,6 +429,12 @@ private:
 
   // FfiClient listener ID (0 means no listener registered)
   int listener_id_{0};
+
+  /// Find a currently subscribed remote track matching the given participant
+  /// identity and track name. Returns nullptr if no such subscribed track
+  /// exists. Acquires @ref lock_.
+  std::shared_ptr<Track> findSubscribedRemoteTrack(const std::string& participant_identity,
+                                                   const std::string& track_name) const;
 
   void onEvent(const proto::FfiEvent& event);
 
--- a/include/livekit/subscription_thread_dispatcher.h
+++ b/include/livekit/subscription_thread_dispatcher.h
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <atomic>
+#include <cstddef>
 #include <cstdint>
 #include <functional>
 #include <memory>
@@ -65,13 +67,15 @@ using DataFrameCallbackId = std::uint64_t;
 ///
 /// `SubscriptionThreadDispatcher` is the low-level companion to @ref Room's
 /// remote track subscription flow. `Room` forwards user-facing callback
-/// registration requests here, and then calls @ref handleTrackSubscribed and
-/// @ref handleTrackUnsubscribed as room events arrive.
+/// registration requests here. For remote audio and video subscriptions it
+/// calls @ref handleTrackSubscribed and @ref handleTrackUnsubscribed; for
+/// data tracks it calls @ref handleDataTrackPublished and
+/// @ref handleDataTrackUnpublished.
 ///
-/// For each registered `(participant identity, track name)` pair, this class
-/// may create a dedicated @ref AudioStream or @ref VideoStream and a matching
-/// reader thread. That thread blocks on stream reads and invokes the
-/// registered callback with decoded frames.
+/// For each registered audio or video `(participant identity, track name)`
+/// pair, this class may create a dedicated @ref AudioStream or @ref
+/// VideoStream and a matching reader thread. That thread blocks on stream
+/// reads and invokes the registered callback with decoded frames.
 ///
 /// This type is intentionally independent from @ref RoomDelegate. High-level
 /// room events such as `RoomDelegate::onTrackSubscribed()` remain in @ref Room,
@@ -89,40 +93,55 @@ public:
   /// Stops all active readers and clears all registered callbacks.
   ~SubscriptionThreadDispatcher();
 
-  /// Register or replace an audio frame callback for a remote subscription.
+  /// Register an audio frame callback for a remote subscription.
   ///
   /// The callback is keyed by remote participant identity plus @p track_name.
   /// If the matching remote audio track is already subscribed, @ref Room may
   /// immediately call @ref handleTrackSubscribed to start a reader.
+  ///
+  /// Registration only succeeds when no reader is currently active for the
+  /// key. To replace a callback whose reader is already running, call
+  /// @ref clearOnAudioFrameCallback first, then register again.
   ///
   /// @param participant_identity Identity of the remote participant.
   /// @param track_name           Track name to match.
   /// @param callback             Function invoked for each decoded audio frame.
   /// @param opts                 Options used when creating the backing
   ///                             @ref AudioStream.
-  void setOnAudioFrameCallback(const std::string& participant_identity, const std::string& track_name,
-                               AudioFrameCallback callback, const AudioStream::Options& opts = {});
+  /// @return @c true if the callback was registered; @c false if a reader is
+  ///         already active for the key (the registration is left unchanged).
+  [[nodiscard]] bool trySetOnAudioFrameCallback(const std::string& participant_identity, const std::string& track_name,
+                                                AudioFrameCallback callback, const AudioStream::Options& opts = {});
 
-  /// Register or replace a video frame callback for a remote subscription.
+  /// Register a video frame callback for a remote subscription.
   ///
   /// The callback is keyed by remote participant identity plus @p track_name.
   /// If the matching remote video track is already subscribed, @ref Room may
   /// immediately call @ref handleTrackSubscribed to start a reader.
+  ///
+  /// Registration only succeeds when no reader is currently active for the
+  /// key. To replace a callback whose reader is already running, call
+  /// @ref clearOnVideoFrameCallback first, then register again.
   ///
   /// @param participant_identity Identity of the remote participant.
   /// @param track_name           Track name to match.
   /// @param callback             Function invoked for each decoded video frame.
   /// @param opts                 Options used when creating the backing
   ///                             @ref VideoStream.
-  void setOnVideoFrameCallback(const std::string& participant_identity, const std::string& track_name,
-                               VideoFrameCallback callback, const VideoStream::Options& opts = {});
+  /// @return @c true if the callback was registered; @c false if a reader is
+  ///         already active for the key (the registration is left unchanged).
+  [[nodiscard]] bool trySetOnVideoFrameCallback(const std::string& participant_identity, const std::string& track_name,
+                                                VideoFrameCallback callback, const VideoStream::Options& opts = {});
 
-  /// Register or replace a rich video frame event callback for a remote
-  /// subscription.
+  /// Register a rich video frame event callback for a remote subscription.
   ///
   /// The callback is keyed by remote participant identity plus @p track_name.
   /// If the matching remote video track is already subscribed, @ref Room may
   /// immediately call @ref handleTrackSubscribed to start a reader.
+  ///
+  /// Registration only succeeds when no reader is currently active for the
+  /// key. To replace a callback whose reader is already running, call
+  /// @ref clearOnVideoFrameCallback first, then register again.
   ///
   /// @param participant_identity Identity of the remote participant.
   /// @param track_name           Track name to match.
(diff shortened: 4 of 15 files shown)


src/room.cpp
```diff
@@ -370,28 +370,95 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  }
+  auto track = findSubscribedRemoteTrack(participant_identity, track_name);
+  if (!track) {
+    LK_LOG_ERROR("Room::trySetOnAudioFrameCallback: track not found for participant={} track_name={}",
+                 participant_identity, track_name);
+    return false;
+  }
+  subscription_thread_dispatcher_->handleTrackSubscribed(participant_identity, track_name, track);
+  return true;
+}
```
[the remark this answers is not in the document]
stephen-derosa: no longer relevant

src/room.cpp
```diff
@@ -370,28 +370,95 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  }
+  auto track = findSubscribedRemoteTrack(participant_identity, track_name);
+  if (!track) {
+    LK_LOG_ERROR("Room::trySetOnVideoFrameEventCallback: track not found for participant={} track_name={}",
+                 participant_identity, track_name);
+    return false;
+  }
+  subscription_thread_dispatcher_->handleTrackSubscribed(participant_identity, track_name, track);
+  return true;
+}
```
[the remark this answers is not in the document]
stephen-derosa: no longer relevant

src/room.cpp
```diff
@@ -370,28 +370,95 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  (void)trySetOnVideoFrameCallback(participant_identity, track_name, std::move(callback), opts);
 }
 
 void Room::setOnVideoFrameEventCallback(const std::string& participant_identity, const std::string& track_name,
                                         VideoFrameEventCallback callback, const VideoStream::Options& opts) {
-  if (subscription_thread_dispatcher_) {
-    subscription_thread_dispatcher_->setOnVideoFrameEventCallback(participant_identity, track_name, std::move(callback),
-                                                                  opts);
-  }
+  (void)trySetOnVideoFrameEventCallback(participant_identity, track_name, std::move(callback), opts);
```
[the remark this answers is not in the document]
stephen-derosa: added a log here instead. Replace functionality never actually worked so this isnt really "breaking"

src/room.cpp
```diff
@@ -370,28 +370,95 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  if (!subscription_thread_dispatcher_->trySetOnAudioFrameCallback(participant_identity, track_name,
+                                                                   std::move(callback), opts)) {
+    return false;
+  }
+  auto track = findSubscribedRemoteTrack(participant_identity, track_name);
+  if (!track) {
+    LK_LOG_ERROR("Room::trySetOnAudioFrameCallback: track not found for participant={} track_name={}",
+                 participant_identity, track_name);
+    return false;
+  }
```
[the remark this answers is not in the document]
stephen-derosa: no longer relevant

src/room.cpp
```diff
@@ -370,28 +370,95 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  auto track = findSubscribedRemoteTrack(participant_identity, track_name);
+  if (!track) {
+    LK_LOG_ERROR("Room::trySetOnVideoFrameCallback: track not found for participant={} track_name={}",
+                 participant_identity, track_name);
+    return false;
+  }
+  subscription_thread_dispatcher_->handleTrackSubscribed(participant_identity, track_name, track);
+  return true;
+}
```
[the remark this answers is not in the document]
stephen-derosa: no longer relevant

src/subscription_thread_dispatcher.cpp
```diff
@@ -581,13 +662,21 @@ std::thread SubscriptionThreadDispatcher::startDataReaderLocked(DataFrameCallbac
...
       const std::scoped_lock<std::mutex> guard(reader->sub_mutex);
+      // A replacement or teardown may have cancelled this reader while the
+      // subscribe was in flight. Close the fresh stream and bail so we do not
+      // leave a second live subscription behind.
+      if (reader->cancelled.load()) {
+        stream->close();
+        return;
+      }
       reader->stream = stream;
     }
```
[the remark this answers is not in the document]
stephen-derosa: The orphaned-thread scenario isn't reachable — every site that sets cancelled also erases the slot and extracts the thread under lock_, so the early return has nothing joinable left in the map. But its right that the cancelled path _should_ mirror the normal-exit `eraseDataReaderIfCurrent` cleanup for symmetry/robustness; I've added it after releasing sub_mutex to keep the lock_ → sub_mutex order and avoid the inversion you flagged.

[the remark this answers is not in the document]
stephen-derosa: done

src/subscription_thread_dispatcher.cpp
```diff
@@ -103,6 +128,26 @@ void SubscriptionThreadDispatcher::setOnVideoFrameCallback(const std::string& pa
...
+                                                                VideoFrameEventCallback callback,
+                                                                const VideoStream::Options& opts) {
+  (void)trySetOnVideoFrameEventCallback(participant_identity, track_name, std::move(callback), opts);
+}
+
+void SubscriptionThreadDispatcher::setOnVideoFrameCallback(const std::string& participant_identity,
+                                                           const std::string& track_name, VideoFrameCallback callback,
+                                                           const VideoStream::Options& opts) {
+  (void)trySetOnVideoFrameCallback(participant_identity, track_name, std::move(callback), opts);
 }
```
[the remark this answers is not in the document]
stephen-derosa: no longer relevant

src/room.cpp
```diff
@@ -370,27 +370,124 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
                                         VideoFrameEventCallback callback, const VideoStream::Options& opts) {
-  if (subscription_thread_dispatcher_) {
-    subscription_thread_dispatcher_->setOnVideoFrameEventCallback(participant_identity, track_name, std::move(callback),
-                                                                  opts);
+  bool result = trySetOnVideoFrameEventCallback(participant_identity, track_name, std::move(callback), opts);
+  if (!result) {
+    LK_LOG_ERROR("Room::setOnVideoFrameEventCallback: failed to set callback for participant={} track_name={}",
+                 participant_identity, track_name);
   }
 }
```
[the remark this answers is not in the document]
stephen-derosa: this was never _really_ being done so its not a true regression

include/livekit/remote_data_track.h
```diff
@@ -106,8 +101,11 @@ class RemoteDataTrack {
 
 private:
   friend class Room;
+#ifdef LIVEKIT_TEST_ACCESS
```
alan-george-lk: Not sure if better or worse, but I typically use GTest's `FRIEND_TEST` macro so you can just give a single test suite/case access to what it needs. May break down a bit between unit and integration test binaries if they have different suite names (could unify name). Basically decays down to a `friend class <blah>` anyways, just looks a bit cleaner: https://google.github.io/googletest/reference/testing.html#FRIEND_TEST

include/livekit/room.h
```diff
@@ -303,16 +303,92 @@ class LIVEKIT_API Room {
...
+  ///
+  /// @param participant_identity Identity of the remote participant.
+  /// @param track_name           Track name to match.
+  /// @param callback             Function invoked for each decoded audio frame.
+  /// @param opts                 Options used when creating the backing
+  ///                             @ref AudioStream.
+  /// @return @c true if the callback was registered; @c false if a reader is
+  ///         already active for the key (call @ref clearOnAudioFrameCallback
+  ///         first) or the room has no dispatcher.
+  [[nodiscard]] bool trySetOnAudioFrameCallback(const std::string& participant_identity, const std::string& track_name,
```
alan-george-lk, line 329: From what I can tell, these new "try" methods have the same inputs as the deprecated prior ones, just a new return type (and name).

This is actually a situation where I think maintaining the API and using exceptions is cleaner than deprecation/new API just to get the `bool` return type.

Thoughts on this? Could we not move the new implementations from `room.cpp` into the older versions and throw instead of returning false?

stephen-derosa: isnt that still changing the public API though? maybe im wrong but i would think adding a throw to a function that previously didnt would require a major bump?

alan-george-lk: Fair point, _technically_ speaking those methods were never marked `noexcept` and anything within them could throw (STL container exceptions, runtime exceptions from FFI stuff, etc.) so I think it's the least of the evils?

include/livekit/room.h
```diff
@@ -303,16 +303,92 @@ class LIVEKIT_API Room {
...
+  ///
+  /// @param participant_identity Identity of the remote participant.
+  /// @param track_name           Track name to match.
+  /// @param callback             Function invoked for each decoded video frame.
+  /// @param opts                 Options used when creating the backing
+  ///                             @ref VideoStream.
+  /// @return @c true if the callback was registered; @c false if a reader is
+  ///         already active for the key (call @ref clearOnVideoFrameCallback
+  ///         first) or the room has no dispatcher.
+  [[nodiscard]] bool trySetOnVideoFrameCallback(const std::string& participant_identity, const std::string& track_name,
```
xianshijing-lk, line 349: I guess I didn't understand why these callback events need to be in the room ? 
couldn't they be in the remote participants? 

And does this frame callbacks work on both local / remote participants ?

src/room.cpp
```diff
@@ -370,27 +370,124 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  }
+
+  // If we've already subscribed to the track, handle it immediately
+  auto track = findSubscribedRemoteTrack(participant_identity, track_name);
+  if (track) {
+    subscription_thread_dispatcher_->handleTrackSubscribed(participant_identity, track_name, track);
+  } else {
+    // The track is not subscribed yet. The callback is registered; the reader
+    // starts when the track is subscribed (see kTrackSubscribed in onEvent).
+    LK_LOG_DEBUG(
```
xianshijing-lk, line 437: should you still return true here ?

src/room.cpp
```diff
@@ -370,27 +370,124 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  }
+  if (!subscription_thread_dispatcher_->trySetOnAudioFrameCallback(participant_identity, track_name,
+                                                                   std::move(callback), opts)) {
+    return false;
+  }
+
+  // If we've already subscribed to the track, handle it immediately
+  auto track = findSubscribedRemoteTrack(participant_identity, track_name);
+  if (track) {
+    subscription_thread_dispatcher_->handleTrackSubscribed(participant_identity, track_name, track);
```
xianshijing-lk, line 433: I think I forgot some technical details here, is this subscription_thread_dispatcher_ a manager that manages all the subscription thread ? 
and does one callback has one corresponding subscription thread ? or one thread that handles all the callbacks ?

src/room.cpp
```diff
@@ -370,27 +370,124 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
 void Room::setOnVideoFrameCallback(const std::string& participant_identity, const std::string& track_name,
                                    VideoFrameCallback callback, const VideoStream::Options& opts) {
-  if (subscription_thread_dispatcher_) {
-    subscription_thread_dispatcher_->setOnVideoFrameCallback(participant_identity, track_name, std::move(callback),
-                                                             opts);
+  bool const result = trySetOnVideoFrameCallback(participant_identity, track_name, std::move(callback), opts);
+  if (!result) {
+    LK_LOG_ERROR("Room::setOnVideoFrameCallback: failed to set callback for participant={} track_name={}",
+                 participant_identity, track_name);
   }
```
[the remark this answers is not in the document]
stephen-derosa: used a log that this is deprecated

include/livekit/room.h
stephen-derosa: the room has access to the room events which are needed to set callbacks to incoming track messages -- the remote participant doesnt really.

This callback only works for local participants, the intention is to provide a nice ergonomic way for the user to set callbacks for _incoming_ messages.

src/room.cpp
stephen-derosa: good question, yes since the user has successfully _set_ the callback. It doesnt necessarily mean that a track has been published or that they will get messages on said track. If we were to only return `true` to a callback being set _after_ a track has been published, then early joining participants that want incoming messages will effectively have to periodically set callbacks until success, or listen to room events -- both of which defeat the purpose of the ergonomic functionality of these setOn*Callback functions.

src/room.cpp
stephen-derosa: > I think I forgot some technical details here, is this subscription_thread_dispatcher_ a manager that manages all the subscription thread ?

it manages subscribing to tracks and providing incoming frames to the users callbacks

> and does one callback has one corresponding subscription thread ? or one thread that handles all the callbacks ?

Each subscription gets its own thread

include/livekit/remote_data_track.h
stephen-derosa: nice much cleaner -- ideal to _not_ introduce a new macro

include/livekit/room.h
stephen-derosa: hm yeah im a little torn here. I dont want to just start throwing underneath applications, especially when our mantra has _somewhat_ been team #nothrow 😅 

but the point of _other_ things can throw inside the function are valid 🤔

include/livekit/subscription_thread_dispatcher.h
stephen-derosa, line 88: @alan-george-lk 
In this work, i realized `subscription_thread_manager` is marked public:

I missed this in the review for the LIVEKIT_API PR. any thoughts on making this private?

I ran a query on this and it doesnt look like:
1. this file is actually included in any of the other public includes
2. the SubscriptionThreadDispatcher doesnt really work without the room wrapper
3. its only used by the room

include/livekit/room.h
alan-george-lk: Also torn, I don't love my suggestion, to me keeping the API clean/avoiding deprecations is the least of the evils. Maybe @xianshijing-lk can chime in

include/livekit/subscription_thread_dispatcher.h
stephen-derosa: discussed in the standup need to confirm that:
- cpp examples build with this private
- it is in fact not included/usable outside the room.cpp/hpp

stephen-derosa:
ensure that this also works for local publishing of video/audio

include/livekit/subscription_thread_dispatcher.h
stephen-derosa: consider moving to src/ so that this isn't reachable

include/livekit/subscription_thread_dispatcher.h
```diff
@@ -95,6 +99,20 @@ class LIVEKIT_API SubscriptionThreadDispatcher {
   /// If the matching remote audio track is already subscribed, @ref Room may
   /// immediately call @ref handleTrackSubscribed to start a reader.
   ///
+  /// Registering again for a key that already has an active reader replaces the
+  /// callback in place: the previous reader's stream is closed and its thread
+  /// is joined before this call returns, and @ref Room then starts a fresh
+  /// reader bound to the new callback. When this call returns, the previous
```
stephen-derosa, line 105: at first i thought this was a bit verbose, but i actually think this is appropriate

include/livekit/subscription_thread_dispatcher.h
stephen-derosa: keeping as public but marking as deprecated to avoid major version bump

include/livekit/subscription_thread_dispatcher.h
```diff
@@ -357,6 +390,13 @@ class LIVEKIT_API SubscriptionThreadDispatcher {
   /// Active stream/thread state keyed by @ref CallbackKey.
   std::unordered_map<CallbackKey, ActiveReader, CallbackKeyHash> active_readers_;
 
+  /// Currently subscribed remote audio/video tracks keyed by @ref CallbackKey.
+  std::unordered_map<CallbackKey, std::shared_ptr<Track>, CallbackKeyHash> subscribed_tracks_;
```
xianshijing-lk, line 394: will this create races or confusion to our users regarding the subscribed tracks ? 

I think our room has API about RemoteTrackPublication, that tracks whether a publication is subscribed. And this |subscribed_tracks_| might confuse developers on which one is the ground truth regarding the subscribed tracks ? 

or are they strictly in sync?

include/livekit/room.h
```diff
@@ -313,16 +316,39 @@ class LIVEKIT_API Room {
...
+  /// @brief Sets the callback for frames from a remote audio track.
+  ///
+  /// The callback can be set before or after the matching track is subscribed
+  /// and runs on a dedicated reader thread.
+  ///
+  /// @param participant_identity Identity of the remote participant.
+  /// @param track_name Name of the remote audio track.
+  /// @param callback Function invoked for each decoded audio frame.
+  /// @param opts Options used to create the backing audio stream.
   void setOnAudioFrameCallback(const std::string& participant_identity, const std::string& track_name,
```
xianshijing-lk, line 328: Do we have doc or comment to explain that when a callback approach is being used, AudioStream / VideoStream APIs can't be used at the same time ? 

Not sure if this applies to DataStream (if we have it)

include/livekit/subscription_thread_dispatcher.h
```diff
@@ -16,17 +16,17 @@
 
 #pragma once
```
xianshijing-lk, line 17: do we have plans to make this subscription_thread_dispatcher internal for the future?  

I think this class has too many implementation details and might confuse our developers.

src/room.cpp
```diff
@@ -400,27 +401,78 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+  return nullptr;
+}
+
 void Room::setOnAudioFrameCallback(const std::string& participant_identity, const std::string& track_name,
                                    AudioFrameCallback callback, const AudioStream::Options& opts) {
-  if (subscription_thread_dispatcher_) {
-    subscription_thread_dispatcher_->setOnAudioFrameCallback(participant_identity, track_name, std::move(callback),
-                                                             opts);
+  if (!subscription_thread_dispatcher_) {
+    LK_LOG_ERROR("Room::setOnAudioFrameCallback: subscription_thread_dispatcher_ is nullptr");
```
xianshijing-lk, line 423: nit, can you add an debug assert here ? so that developers know the problem rather than just fail silently ?

xianshijing-lk: or maybe just repalce it with 
assert(subscription_thread_dispatcher_ &&
          "Room::setOnAudioFrameCallback: subscription_thread_dispatcher_ is nullptr");

src/room.cpp
```diff
@@ -400,27 +401,78 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+        participant_identity, track_name);
   }
 }
 
 void Room::setOnVideoFrameCallback(const std::string& participant_identity, const std::string& track_name,
                                    VideoFrameCallback callback, const VideoStream::Options& opts) {
-  if (subscription_thread_dispatcher_) {
-    subscription_thread_dispatcher_->setOnVideoFrameCallback(participant_identity, track_name, std::move(callback),
-                                                             opts);
+  if (!subscription_thread_dispatcher_) {
```
xianshijing-lk, line 442: ditto

src/room.cpp
```diff
@@ -400,27 +401,78 @@ void Room::unregisterByteStreamHandler(const std::string& topic) {
...
+        participant_identity, track_name);
   }
 }
 
 void Room::setOnVideoFrameEventCallback(const std::string& participant_identity, const std::string& track_name,
                                         VideoFrameEventCallback callback, const VideoStream::Options& opts) {
-  if (subscription_thread_dispatcher_) {
-    subscription_thread_dispatcher_->setOnVideoFrameEventCallback(participant_identity, track_name, std::move(callback),
-                                                                  opts);
+  if (!subscription_thread_dispatcher_) {
```
xianshijing-lk, line 461: ditto

src/subscription_thread_dispatcher.cpp
```diff
@@ -57,52 +57,123 @@ SubscriptionThreadDispatcher::~SubscriptionThreadDispatcher() {
...
+    // The caller IS this reader so this function was called from the set callback.
+    // Joining here would be a self-join. Detaching is safe: no reader lambda captures `this`; each
+    // owns its stream, callback, and per-reader state by value, so once
+    // extracted the thread touches nothing owned by the dispatcher.
+    LK_LOG_WARN(
+        "{} was called from inside the frame callback of the reader it stops; "
+        "detaching that reader instead of self-joining. It exits once the "
+        "callback returns. Registering, clearing, or tearing down from within a "
+        "frame callback is discouraged",
+        operation);
```
xianshijing-lk, line 78: so operation is used for logging only ?

src/subscription_thread_dispatcher.cpp
```diff
@@ -57,52 +57,123 @@ SubscriptionThreadDispatcher::~SubscriptionThreadDispatcher() {
...
+}
+
+void SubscriptionThreadDispatcher::finishReaderDrainAndRestart(const CallbackKey& key, std::thread old_thread,
+                                                               const char* operation) {
+  const bool drained = old_thread.joinable();
+  disposeReaderThread(std::move(old_thread), operation);
+
+  const std::scoped_lock<std::mutex> lock(lock_);
+  if (drained) {
+    auto it = draining_readers_.find(key);
```
xianshijing-lk, line 100: just want to make sure there can be only one entry in draining_readers_ per key, right ? the question is mostly for data track readers ?

src/subscription_thread_dispatcher.cpp
```diff
@@ -57,52 +57,123 @@ SubscriptionThreadDispatcher::~SubscriptionThreadDispatcher() {
...
-      opts,
-  };
-  LK_LOG_DEBUG(
-      "Registered video frame event callback for participant={} track_name={} "
-      "replacing_existing={} total_video_callbacks={}",
-      participant_identity, track_name, replacing, video_callbacks_.size());
+  std::thread old_thread;
+  {
+    const std::scoped_lock<std::mutex> lock(lock_);
+    old_thread = extractReaderForDrainLocked(key);
```
xianshijing-lk, line 142: curiously, for the first setOnVideoFrameEventCallback() call, should old_thread exist  ? 
or we create one thread as old_thread ? which is not joinable ?

src/subscription_thread_dispatcher.cpp
```diff
@@ -606,9 +742,14 @@ std::thread SubscriptionThreadDispatcher::startDataReaderLocked(DataFrameCallbac
...
           track_name, identity, static_cast<std::uint32_t>(error->code), error->message);
     }
+    // Whether the stream ended on its own (server EOS) or was closed by an
+    // extract/teardown, record that this reader is done so a slot still
+    // holding it is treated as replaceable.
+    markDataReaderFinished(reader);
     LK_LOG_INFO("Data reader thread exiting for \"{}\" track=\"{}\"", identity, track_name);
   });
   // NOLINTEND(bugprone-lambda-function-name)
+  reader->thread_id = reader->thread.get_id();
```
xianshijing-lk, line 752: curiously, what this thread_id is used for ? I only see code that get it, but not use it, maybe I miss some code.

README.md
```diff
@@ -191,6 +191,31 @@ room->addOnDataFrameCallback(sender_identity, "app-data",
...
+subscribed. If the track is already subscribed, a reader starts immediately;
+otherwise it starts when the subscription arrives.
+
+Calling `setOnAudioFrameCallback` / `setOnVideoFrameCallback` /
+`setOnVideoFrameEventCallback` again for the same
+`(participant_identity, track_name)` **replaces** the callback in place. The
+previous reader is stopped and its thread joined, and only then is a fresh reader
+started bound to the new callback — the old and new callbacks never run at the
+same time, and there is no need to call `clearOn*FrameCallback` first. Two
+consequences worth knowing:
```
xianshijing-lk, line 204: probably here is the place we should mention those setOnXXXCallback can't be used together with AudioStream/VideoStream/DataStream, which could create races

src/subscription_thread_dispatcher.cpp
```diff
@@ -57,52 +57,132 @@ SubscriptionThreadDispatcher::~SubscriptionThreadDispatcher() {
...
+    const bool replacing = audio_callbacks_.find(key) != audio_callbacks_.end();
+    audio_callbacks_[key] = RegisteredAudioCallback{std::move(callback), opts};
+    LK_LOG_DEBUG(
+        "Registered audio frame callback for participant={} track_name={} "
+        "replacing_existing={} stopped_reader={} total_audio_callbacks={}",
+        participant_identity, track_name, replacing, old_thread.joinable(), audio_callbacks_.size());
+  }
+  // Joins the previous reader first, then starts a fresh one bound to the new
+  // callback if the track is subscribed.
+  finishReaderDrainAndRestart(key, std::move(old_thread), "setOnAudioFrameCallback");
```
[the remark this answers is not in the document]
xianshijing-lk: This seems to a valid race, what do you think ?

PR #198, opened 2026-07-07
Add degradation preference options

alan-george-lk:
Summary: Expose `DegradationPreference` on `TrackPublishOptions `and wire it through the FFI publish path

- Add `DegradationPreference` and `TrackPublishOptions::degradation_preference`
- Serialize/deserialize the field in `room_proto_converter.cpp`
- Add unit coverage for defaults and proto round-trip

Misc changes found to be needed during development:
- Make failIfNotConfigured() throw so missing test env vars fail fast
- In `AGENTS.md` document that test tokens must be sourced before running integration/stress tests

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -352,10 +352,20 @@ CI step will fail loudly if its symbols escape the public ABI of
 
 Tests are under `src/tests/` using Google Test:
 
-```
+```bash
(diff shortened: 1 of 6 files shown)


xianshijing-lk (approved):
lgtm, thanks for taking care of it

PR #202, opened 2026-07-08
feat: Add missing reply_to_stream_id field to TextStreamInfo

Soralsei:
Add missing `reply_to_stream_id` field to `TextStreamInfo` and forward it from `proto::DataStream.text_header().reply_to_stream_id()` in `room_proto_converter.cpp:makeTextInfo(...)`

### DIFF
--- a/include/livekit/data_stream.h
+++ b/include/livekit/data_stream.h
@@ -64,7 +64,7 @@ struct TextStreamInfo : BaseStreamInfo {
   /// IDs of any attached streams (for attached files).
   std::vector<std::string> attachments;
   /// If this stream is a reply to another stream, this field holds its ID
-  std::optional<std::string> reply_to_stream_id = std::nullopt;
+  std::optional<std::string> reply_to_stream_id;
 };
 
 /// Metadata for a byte stream.
--- a/include/livekit/data_stream.h
+++ b/include/livekit/data_stream.h
@@ -61,8 +61,10 @@ struct BaseStreamInfo {
 
 /// Metadata for a text stream.
 struct TextStreamInfo : BaseStreamInfo {
-  /// IDs of any attached streams (for replies / threads).
+  /// IDs of any attached streams (for attached files).
   std::vector<std::string> attachments;
+  /// If this stream is a reply to another stream, this field holds its ID
+  std::optional<std::string> reply_to_stream_id = std::nullopt;
 };
 
 /// Metadata for a byte stream.
--- a/src/tests/integration/test_data_streams.cpp
+++ b/src/tests/integration/test_data_streams.cpp
@@ -81,7 +81,7 @@ public:
   const std::string& senderIdentity() const { return sender_identity_; }
 
 private:
-  std::thread recv_thread_{};
+  std::thread recv_thread_;
   std::mutex mutex_;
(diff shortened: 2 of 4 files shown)


alan-george-lk:
Hey @Soralsei, thanks for the contribution! Will get back to you ASAP on feedback. We have some instability right now with PlatformAudio tanking CI runs, please re-run your actions when convenient as needed (sorry, working on that fix as well).

include/livekit/data_stream.h
```diff
@@ -61,8 +61,10 @@ struct BaseStreamInfo {
 
 /// Metadata for a text stream.
 struct TextStreamInfo : BaseStreamInfo {
-  /// IDs of any attached streams (for replies / threads).
+  /// IDs of any attached streams (for attached files).
   std::vector<std::string> attachments;
+  /// If this stream is a reply to another stream, this field holds its ID
+  std::optional<std::string> reply_to_stream_id = std::nullopt;
```
alan-george-lk: Nit: repo convention (and also elsewhere in this file) is to not default assign `std::nullopt`, is a bit redundant

src/tests/integration/test_data_streams.cpp
```diff
@@ -0,0 +1,226 @@
...
+  const auto sender_identity = lockLocalParticipant(*sender_room)->identity();
+
+  ByteStreamCollector collector;
+  collector.registerOn(*receiver_room, topic);
+
+  {
+    ByteStreamWriter writer(*lockLocalParticipant(*sender_room), /*name=*/"payload.bin", topic);
+    writer.write(payload);
+    writer.close();
+  }
```
alan-george-lk: Can you explain why this is scoped like this?

Also consider locking the local participant once in this test, unless there's some logical reason not to

src/tests/integration/test_data_streams.cpp
```diff
@@ -0,0 +1,226 @@
...
+    std::unique_lock<std::mutex> lock(mutex_);
+    return cv_.wait_for(lock, timeout, [this] { return done_; });
+  }
+
+  const ByteStreamInfo& info() const { return info_; }
+  const std::vector<std::uint8_t>& content() const { return content_; }
+  const std::string& senderIdentity() const { return sender_identity_; }
+
+private:
+  std::thread recv_thread_{};
```
alan-george-lk: Nit: can drop `{}` here, and we tend to prefer initializer lists in constructors vs. inline

alan-george-lk (commented):
Overall looks good: just a few nitpicks and curiosity comments to address

include/livekit/data_stream.h
stephen-derosa: +1, also aligns with the rest of the code base

stephen-derosa (commented):
overall looks pretty good, one small comment

src/tests/integration/test_data_streams.cpp
Soralsei: No particular reason except visual separation between sender/receiver logic. Removed in the latest commit

alan-george-lk (approved):

alan-george-lk:
Thanks for the contribution -- this should be available in the next SDK release.

PR #205, opened 2026-07-13
Bugfix: Server-initiated room disconnect shutdown fix

alan-george-lk:
## Summary
- Fixes server-initiated room disconnect shutdown by routing explicit disconnect, destructor cleanup, server disconnect events, and EOS through a shared `shutdown()` helper, which by proxy cleans up event code
- Ensure server Disconnected events remove the C++ FFI listener, stop subscriptions, clear room-owned state, and fire onDisconnected exactly once without re-sending an FFI disconnect

Before fix, when server deleted the room:

```log
[2026-07-13 10:02:31.515] [livekit] [error] FfiClient listener threw: mutex lock failed: Invalid argument
```

Above observed on Mac, but crashed on Linux (user-reported). No longer after the fix.

## Testing
- Original issue reproduced via local binary (not committed), then fixed and confirmed no longer occurred
- Add regression coverage for synthetic FFI server disconnects and local-SFU room deletion via `lk --dev room delete`
- Consolidates `emitFfiEvent` into standalone header helper `ffi_utils.h` for use across two tests now

### Ticket
BOT-464


include/livekit/room.h
```diff
@@ -355,5 +355,10 @@ class LIVEKIT_API Room {
   int listener_id_{0};
 
   void onEvent(const proto::FfiEvent& event);
+
+  // Tracks local teardown independently from connection_state_ to handle all teardown cases.
+  bool teardown_started_{false};
+  // Shared teardown path for explicit disconnect, server disconnect, EOS, and destruction.
+  bool teardown(bool disconnect_ffi, DisconnectReason reason, bool notify_delegate);
```
xianshijing-lk: nit, is shutdown a better name ? I think we use shutdown in some other classes

src/room.cpp
```diff
@@ -224,58 +228,81 @@ bool Room::disconnect(DisconnectReason reason) {
 
   {
     const std::scoped_lock<std::mutex> g(lock_);
-    if (connection_state_ == ConnectionState::Disconnected) {
-      // Already torn down (or never connected). Nothing to do.
+    const bool has_room_state = connection_state_ != ConnectionState::Disconnected || listener_id_ != 0 ||
+                                room_handle_ || local_participant_ || !remote_participants_.empty();
+    if (teardown_started_ || !has_room_state) {
       return false;
```
xianshijing-lk, line 236: should we return true if nothing to do or it is being torndown ?

src/tests/unit/test_room.cpp
```diff
@@ -33,6 +65,48 @@ class RoomTest : public ::testing::Test {
...
+    reason = event.reason;
+  }
+
+  int count = 0;
+  DisconnectReason reason = DisconnectReason::Unknown;
+};
+
+} // namespace
+
+TEST_F(RoomTest, ServerDisconnectTearsDownRoomAndRemovesListener) {
```
xianshijing-lk: can you add a test that reconnect the room after teardown ?

xianshijing-lk (approved):
lgtm, assuming you address those comments.

include/livekit/room.h
alan-george-lk: Agreed — renamed all teardown references to shutdown

src/room.cpp
alan-george-lk: I'm open to changing this, right now our public API for `disconnect()` states:

> `@returns true` if the graceful disconnect succeeds; `false` if the room was already disconnected (no-op) or the graceful disconnect fails.

Which this matches, `false` == noop. I get how that could be confusing, but I think the idea is `true` is when the disconnect actually happened

src/tests/unit/test_room.cpp
alan-george-lk: Added

src/room.cpp
xianshijing-lk: That is fair. We can keep the current behavior to avoid breaking changes.

xianshijing-lk:
lgtm, thanks

src/room.cpp
```diff
@@ -1208,56 +1221,7 @@ void Room::onEvent(const FfiEvent& event) {
...
-          if (listener_to_remove != 0) {
-            FfiClient::instance().removeListener(listener_to_remove);
-          }
-
-          if (old_local_participant) {
-            old_local_participant->shutdown();
-          }
-
-          // old_* state is destroyed here when going out of scope
+          (void)shutdown(false, DisconnectReason::Unknown, false);
```
stephen-derosa, line 1239: is EOS not a DisconnectReason?

src/room.cpp
```diff
@@ -1168,22 +1195,8 @@ void Room::onEvent(const FfiEvent& event) {
...
-            break;
-          }
-          DisconnectedEvent ev;
-          ev.reason = toDisconnectReason(re.disconnected().reason());
-          if (delegate_snapshot) {
-            delegate_snapshot->onDisconnected(*this, ev);
-          }
+          const auto reason = toDisconnectReason(re.disconnected().reason());
+          (void)shutdown(false, reason, true);
           break;
```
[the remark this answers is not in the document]
stephen-derosa: hard to image apps would be using kEos rather than kDisconnected, i guess something to consider

src/room.cpp
stephen-derosa: FWIW we have a public `connectionState` so apps can check the state before trying to disconnect

stephen-derosa (approved):
nice lgtm!

src/room.cpp
alan-george-lk: TODO: Confirm the flow is this on server disconnect:

- Server disconnect event (do shutdown)
- EOS (don't need shutdown)

Verify that the EOS is double-calling shutdown right now as written (after server close)

src/room.cpp
```diff
@@ -1168,20 +1193,17 @@ void Room::onEvent(const FfiEvent& event) {
...
+            // and notifies the delegate itself. Suppress that duplicate while
+            // passing server-initiated disconnects through unchanged.
+            should_notify = connection_state_ != ConnectionState::Disconnected;
+          }
+          if (should_notify && delegate_snapshot) {
+            DisconnectedEvent ev;
+            ev.reason = toDisconnectReason(re.disconnected().reason());
             delegate_snapshot->onDisconnected(*this, ev);
           }
           break;
```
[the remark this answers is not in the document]
alan-george-lk: Resolved

src/room.cpp
alan-george-lk: Fixed this in Rust, EOS was not being emitted right: https://github.com/livekit/rust-sdks/pull/1259

src/tests/integration/test_room.cpp
```diff
@@ -243,7 +285,100 @@ TEST_F(RoomTest, DestructorDisconnect) {
...
+  DisconnectTrackingDelegate delegate;
+  room.setDelegate(&delegate);
+
+  RoomOptions options;
+  ASSERT_TRUE(room.connect(server_url_, token_, options)) << "connect failed";
+  ASSERT_EQ(room.connectionState(), ConnectionState::Connected);
+  ASSERT_NE(room.localParticipant().lock(), nullptr);
+
+  const std::string delete_command = std::string("lk --dev room delete --yes ") + room_name;
+  ASSERT_EQ(std::system(delete_command.c_str()), 0) << "failed to delete local test room";
```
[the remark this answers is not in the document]
alan-george-lk: Dropped LIVEKIT_ROOM passthrough entirely, risk isn't worth a hardcoded room name in a test

PR #211, opened 2026-07-15
Move RPC one minute integration into stress test

alan-george-lk:
Simple PR to move the one minute RPC integration test into a stress test, shaving one minute off each integration test run (local + CI).

Stress tests will be run nightly in a downstream PR.


stephen-derosa (approved):
Lgtm

PR #213, opened 2026-07-17
Bugfix: AudioFrame input sanitization

alan-george-lk:
`AudioFrame` constructor was not fully input sanitized, leading to potential issues with division by 0 or integer overflows later in the class.


stephen-derosa (approved):
looks good pending windows test ci issues

PR #215, opened 2026-07-20
Temporarily disable platform audio integration tests in CI

alan-george-lk:
Disables platform audio integration tests in CI until instability can be isolated and fixed. Has been exposing larger runtime/state/ownership issues in the Rust SDK which are not resolved. Seems to be crashing unrelated tests frequently.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -275,6 +275,8 @@ jobs:
           set -euo pipefail
           source .token_helpers/set_data_track_test_tokens.bash
           build-release/bin/livekit_integration_tests \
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -275,6 +275,8 @@ jobs:
           set -euo pipefail
           source .token_helpers/set_data_track_test_tokens.bash
           build-release/bin/livekit_integration_tests \
+            # Temporarily disable platform audio integration tests in CI until instability is resolved
```
stephen-derosa: can you put a `# TODO(ticket-for-this-bug)` here please

stephen-derosa (approved):

Issue #212, opened 2026-07-16
Feature request: setVideoDimensions on remote video publications (adaptive subscribe resolution)


jiyeyuran:
## Summary

The JavaScript/TypeScript client SDK exposes `RemoteTrackPublication.setVideoDimensions({ width, height })`, which lets subscribers request a target receive resolution per remote video track from the SFU. The C++ client SDK does not appear to offer an equivalent API today.

We would like parity with the JS SDK so native/C++ subscribers can adapt inbound video resolution to downstream layout needs (multi-party compositing, grid tiles, stage/speaker layouts) without subscribing at full publisher resolution for every track.

## Use case

We run server-side media pipelines that subscribe to LiveKit rooms, composite many participants (grid / speaker layouts), and push to RTMP, HLS, or similar outputs. On our browser-based subscriber we already drive adaptive subscriptions from compositor tile sizes:

```ts
pub.setVideoDimensions({ width, height });
```

When a participant is shown in a small grid cell we request something like 320×180; when they are on stage we request a larger cap (e.g. 640×360 or 1280×720). This materially reduces CPU, bandwidth, and decode cost for N-way rooms.

We are building native pipelines on top of the C++ SDK and need the same capability there.

## Expected API (rough sketch)

Something along the lines of:

```cpp
// On a subscribed remote video track publication
publication-> setVideoDimensions(int width, int height);
```

Semantics should match the JS SDK:

- Request the SFU to deliver video at approximately the given dimensions (not necessarily exact pixels).
- Per-track, independent of other subscribers.
- Safe to call when subscription hints / layout change; idempotent when dimensions are unchanged.
- No-op or clear error when the publication is not subscribed or not a video track.

## Why this matters

Without `setVideoDimensions` (or equivalent), a C++ subscriber tends to receive publisher-native resolution on every track. For 9+ participant grid layouts this means:

- Unnecessary downscale work in the compositor
- Higher SFU egress and client decode load
- More visible dimension churn during simulcast/SVC layer switches

The JS SDK already solves this; native subscribers should not be second-class.

## References

- JS SDK: `RemoteTrackPublication.setVideoDimensions` on subscribed remote video publications.
- Related web SDK surface: subscription / dimension control on `RemoteTrackPublication` (naming may differ slightly by SDK version).

## Environment

- Consumer: server-side subscriber (composite / transcode), not a typical mobile client
- Room sizes: often 4–16 video participants
[...]

PR #219, opened 2026-07-22
Additional RemoteTrackPublication options

alan-george-lk:
Features generated from issue #212 

- Additional `RemoteTrackPublication` options that better match the JS SDK where possible (FFI-supported only): https://github.com/livekit/client-sdk-js/blob/main/src/room/track/RemoteTrackPublication.ts


jiyeyuran:
We’ve verified this functionality in our integration, including `setVideoDimensions` for adaptive subscription, and it works as expected with no issues found. We’d appreciate it if this could be merged soon. Thanks!

include/livekit/remote_track_publication.h
```diff
@@ -32,12 +35,77 @@ class LIVEKIT_API RemoteTrackPublication : public TrackPublication {
...
+  /// This overrides a previous call to setVideoQuality().
+  /// Repeating the current dimensions does not send another request.
+  ///
+  /// @param width Requested maximum width in pixels. Must be greater than zero.
+  /// @param height Requested maximum height in pixels. Must be greater than zero.
+  /// @return True if an update was sent; false if the dimensions are unchanged
+  /// or the publication is not a subscribed video track.
+  /// @throws std::runtime_error if the publication has an invalid FFI handle or
+  /// the FFI request fails.
+  bool setVideoDimensions(std::uint32_t width, std::uint32_t height);
```
stephen-derosa, line 96: `const std::uint32_t width` ?

include/livekit/remote_track_publication.h
```diff
@@ -32,12 +35,77 @@ class LIVEKIT_API RemoteTrackPublication : public TrackPublication {
...
+  /// @brief Requests the maximum simulcast quality to receive.
+  ///
+  /// This overrides a previous call to setVideoDimensions().
+  ///
+  /// @param quality Maximum acceptable simulcast quality.
+  /// @return True if an update was sent; false if the quality is unchanged or
+  /// the publication is not a subscribed, simulcasted video track.
+  /// @throws std::runtime_error if the publication has an invalid FFI handle or
+  /// the FFI request fails.
+  bool setVideoQuality(VideoQuality quality);
```
stephen-derosa, line 82: `const VideoQuality quality` ?

include/livekit/remote_track_publication.h
```diff
@@ -32,12 +35,77 @@ class LIVEKIT_API RemoteTrackPublication : public TrackPublication {
...
+  ///
+  /// Disabling a subscribed publication reduces bandwidth without changing its
+  /// desired subscription state.
+  ///
+  /// @param enabled True to enable media delivery; false to disable it.
+  /// @return True if an update was sent; false if the state is unchanged or
+  /// the publication is not subscribed.
+  /// @throws std::runtime_error if the publication has an invalid FFI handle or
+  /// the FFI request fails.
+  bool setEnabled(bool enabled);
```
stephen-derosa, line 65: `const bool enabled`?

src/remote_track_publication.cpp
```diff
@@ -47,4 +48,120 @@ void RemoteTrackPublication::setSubscribed(bool subscribed) {
...
+  }
+
+  proto::FfiRequest req;
+  auto* msg = req.mutable_enable_remote_track_publication();
+  msg->set_track_publication_handle(static_cast<std::uint64_t>(ffiHandleId()));
+  msg->set_enabled(enabled);
+
+  const proto::FfiResponse resp = FfiClient::instance().sendRequest(req);
+  if (!resp.has_enable_remote_track_publication()) {
+    throw std::runtime_error("RemoteTrackPublication::setEnabled: FFI response missing enabled update result");
```
stephen-derosa, line 70: i get throwing for `ffiHandleId() == 0` but do want to throw if we miss a response? Probably, otherwise we have broken our contract with the ffi?

src/remote_track_publication.cpp
```diff
@@ -47,4 +48,120 @@ void RemoteTrackPublication::setSubscribed(bool subscribed) {
...
+  if (!simulcasted()) {
+    LK_LOG_WARN("RemoteTrackPublication::setVideoQuality ignored for non-simulcast publication {}", sid());
+    return false;
+  }
+  if (!subscribed_ || track() == nullptr) {
+    LK_LOG_WARN("RemoteTrackPublication::setVideoQuality ignored for unsubscribed publication {}", sid());
+    return false;
+  }
+  if (video_preference_ == VideoPreference::QUALITY && video_quality_ == quality) {
+    return false;
```
stephen-derosa, line 91: does this means the quality that is desired is already set? If so, should we instead return true here? logic being they are getting the intended result of the function call.

src/remote_track_publication.cpp
```diff
@@ -47,4 +48,120 @@ void RemoteTrackPublication::setSubscribed(bool subscribed) {
...
+  if (width == 0 || height == 0) {
+    LK_LOG_WARN("RemoteTrackPublication::setVideoDimensions requires non-zero dimensions for publication {}", sid());
+    return false;
+  }
+  if (!subscribed_ || track() == nullptr) {
+    LK_LOG_WARN("RemoteTrackPublication::setVideoDimensions ignored for unsubscribed publication {}", sid());
+    return false;
+  }
+  if (video_preference_ == VideoPreference::DIMENSIONS && width_ == width && height_ == height) {
+    return false;
```
stephen-derosa, line 142: same `return true` comment here

include/livekit/remote_track_publication.h
alan-george-lk: `const` actually doesn't have any affect on by value arguments in declarations, only in definitions

include/livekit/remote_track_publication.h
alan-george-lk: https://github.com/livekit/client-sdk-cpp/pull/219/changes#r3815514686

include/livekit/remote_track_publication.h
alan-george-lk: https://github.com/livekit/client-sdk-cpp/pull/219/changes#r3815514686

src/remote_track_publication.cpp
alan-george-lk: Fair callout, but we have a lot of precedent in the SDK for throwing on missing proto responses, the ones in here follow that

alan-george-lk:
> We’ve verified this functionality in our integration, including `setVideoDimensions` for adaptive subscription, and it works as expected with no issues found. We’d appreciate it if this could be merged soon. Thanks!

Hey @jiyeyuran, I am working on getting this wrapped up and hopefully released soon.

src/remote_track_publication.cpp
alan-george-lk: Right now the API is documented as follows:

>   /// @return True if an update was sent; false if the quality is unchanged or
  /// the publication is not a subscribed, simulcasted video track.
  
There's definitely an argument to `true` if good, `false` otherwise. Open to changing this. Note that `room.disconnect()` returns false if a disconnect signal wasn't actually sent, even if it was disconnected already.

include/livekit/remote_track_publication.h
alan-george-lk: Clang-tidy would flag this if it could be `const`

stephen-derosa: i would think that the definition and declaration should be the same for readability. I checked the definition and it also just uses `bool enabled`. why not use `const bool enabled` to avoid potential bugs in the future like setting `enabled` to a new value?

src/remote_track_publication.cpp
stephen-derosa: since its clearly doc'd i think its fine. This ultimately pushes users to check if they already at desired quality before calling this function which i think is ok.

include/livekit/remote_track_publication.h
alan-george-lk: Fair point and good catch, moved all callouts to `const` in definition and declaration

stephen-derosa (approved):

PR #227, opened 2026-07-30
(WIP) Integrate livekit capture

ladvoc:
Integrates https://github.com/livekit/rust-sdks/pull/1299


src/tests/integration/test_capture_source.cpp
```diff
@@ -0,0 +1,161 @@
...
+      });
+
+  // Publish a track backed by the capture source's RTC video source,
+  // merging application options over the source-derived ones.
+  auto track = LocalVideoTrack::createLocalVideoTrack(track_name,
+                                                      capture->videoSource());
+
+  // Application options are merged in; source-dictated fields win.
+  TrackPublishOptions app_options;
+  app_options.source = TrackSource::SOURCE_CAMERA;
```
ladvoc, line 217: from sync: Should the capture source specify this field?

include/livekit/capture_source.h
```diff
@@ -0,0 +1,234 @@
...
+/// Configuration for the built-in test source producing solid-color frames.
+struct DemoVideoSourceConfig {};
+
+/// Why a capture ended without error.
+enum class CaptureExit {
+  /// Stopped via @ref CaptureSource::stop (or source destruction).
+  Stopped = 0,
+  /// The producer reached the end of its stream.
+  EndOfStream = 1,
+};
```
ladvoc, line 312: from sync: Add panic reason.

.github/workflows/tests.yml
```diff
@@ -221,14 +253,18 @@ jobs:
             libabsl-dev \
             libcurl4-openssl-dev \
             libwayland-dev libdecor-0-dev \
+            libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev \
```
alan-george-lk, line 264: Help me understand the difference in versions between these installs and the one at the top of this file for Windows? `1.0-dev` vs. `1.28.5`? Just making sure no surprises

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+  /// Pre-encoded access units, published as passthrough.
+  Encoded = 1,
+};
+
+/// Bitrate unit expected by a GStreamer encoder property.
+enum class GstreamerBitrateUnit {
+  /// Bits per second.
+  Bps = 0,
+  /// Kilobits per second.
+  Kbps = 1,
```
alan-george-lk, line 65: Do we have any precedence for bit rate enums/units elsewhere? Might be worth consolidating if so

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+};
+
+/// Video resolution in pixels.
+struct CaptureResolution {
+  int width = 0;
+  int height = 0;
+};
```
alan-george-lk, line 72: Similar comment here: https://github.com/livekit/client-sdk-cpp/pull/227/changes#r3845438732

If we have a generic video resolution struct somewhere, a smaller API surface would be ideal

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+
+  /// Unit the property expects.
+  GstreamerBitrateUnit unit = GstreamerBitrateUnit::Bps;
+};
+
+/// Configuration for encoded ingest from a GStreamer pipeline.
+struct GstreamerVideoSourceConfig {
+  /// GStreamer launch description for the encoded producer pipeline.
+  ///
+  /// Must contain `appsink name=lk_appsink`, or leave exactly one encoded
```
alan-george-lk: Consider a `@note` or `@warning` here so it's more flagged in the rendered docs

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+  /// Packed BGR24.
+  Bgr24 = 4,
+  /// Packed YUYV/YUY2.
+  Yuyv = 5,
+  /// Packed UYVY.
+  Uyvy = 6,
+  /// Single-plane 8-bit luma.
+  Grey = 7,
+  /// Encoded MJPEG frames.
+  Mjpeg = 8,
```
alan-george-lk: Definitely a fan of enum styling consistency if we have precedence for `CapitalLetterEnum` style, but wondering if some of these would be all caps, like `RGB24`. I'd default first to our SDK precedent and second to how Rust defines them

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+/// platform-specific, so there is no portable default.
+struct DeviceFormat {
+  /// Frame dimensions.
+  CaptureResolution resolution;
+
+  /// Frame rate in frames per second.
+  std::uint32_t framerate_fps = 0;
+
+  /// Frame format.
+  DeviceFrameFormat frame_format = DeviceFrameFormat::Nv12;
```
alan-george-lk: Should we add a `DeviceFrameFormat::Invalid` or `Unset` to enforce users setting this? Otherwise they could skip this one and the API call would go through. Re doc comment:

> /// All three components are required.

This comment likely applies to the other enums in this file

alan-george-lk: Just saw the comment later in this file:

> /// @ref DeviceFrameFormat::Nv12 is the only value both backends accept, which
/// is why it is @ref DeviceFormat's default.

Maybe the comments just need to be aligned and don't need to add invalid/unset

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+  static DeviceFormatRequest exact(DeviceFormat format);
+
+  /// Use the device's closest supported resolution and frame rate.
+  static DeviceFormatRequest closest(DeviceFormat format);
+
+  /// Prefer the highest frame rate, optionally constrained.
+  static DeviceFormatRequest highestFramerate(HighestFramerateConstraint constraint = {});
+
+  /// Prefer the highest resolution, optionally constrained.
+  static DeviceFormatRequest highestResolution(HighestResolutionConstraint constraint = {});
```
alan-george-lk, line 243: Any reason these each shouldn't have full doc comments? `@param` and `@return`

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+  ///
+  /// The meaning is platform-defined and is not necessarily the position of
+  /// the device in @ref CaptureSource::listDevices: on macOS it indexes the
+  /// AVFoundation video-device list, on Linux it is the V4L2 node number
+  /// (@c /dev/videoN). Either way the value can shift as devices are attached
+  /// and removed. Prefer @ref id, which is stable on macOS.
+  static DeviceSelector index(std::uint32_t device_index);
+
+  /// A platform-stable identifier, as reported by @ref CaptureDeviceInfo::id.
+  static DeviceSelector id(std::string device_id);
```
alan-george-lk, line 279: Same full doc comment question

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+  /// format request, unsupported platform, missing capture feature) are
+  /// thrown from the future as @ref CaptureSourceError.
+  static std::future<std::shared_ptr<CaptureSource>> create(DeviceVideoSourceConfig config);
+
+  /// List the video capture devices available on this machine.
+  ///
+  /// Completes asynchronously: enumeration queries the platform capture stack
+  /// and may block briefly. Throws @ref CaptureSourceError from the future on
+  /// failure, including on platforms without a capture backend.
+  static std::future<std::vector<CaptureDeviceInfo>> listDevices();
```
alan-george-lk, line 415: Same doc comment question

src/capture_source.cpp
```diff
@@ -0,0 +1,459 @@
...
+}
+
+std::future<std::vector<CaptureDeviceInfo>> CaptureSource::listDevices() {
+  // Issuing the request can fail synchronously (an FFI library built without
+  // the capture feature rejects it outright). Report that through the returned
+  // future so every failure reaches the caller as a CaptureSourceError from
+  // exactly one place, as documented.
+  std::optional<std::future<proto::CaptureDeviceList>> devices;
+  try {
+    devices.emplace(FfiClient::instance().listCaptureDevicesAsync());
```
alan-george-lk: Not really sure what the optional is getting us here, it seems to be emplaced into and then immediately dereferenced on line `318`, with no `has_value` check, not that I think it even can be empty? It's either set or it throws on `311`

src/capture_source.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+std::future<std::shared_ptr<CaptureSource>> CaptureSource::createFromRequest(proto::NewCaptureSourceRequest request) {
+  // See listDevices(): a synchronous send failure is delivered through the
+  // future rather than thrown from the factory.
+  std::optional<std::future<proto::OwnedCaptureSource>> owned;
+  try {
+    owned.emplace(FfiClient::instance().newCaptureSourceAsync(std::move(request)));
+  } catch (const std::exception& e) {
+    return readyCaptureError<std::shared_ptr<CaptureSource>>(e);
+  }
```
alan-george-lk: Same optional comment as https://github.com/livekit/client-sdk-cpp/pull/227/changes#r3845867490

src/capture_source.cpp
```diff
@@ -0,0 +1,459 @@
...
+  // helper thread keeps the returned future's wait semantics standard.
+  return std::async(std::launch::async, [devices = std::move(*devices)]() mutable {
+    try {
+      const proto::CaptureDeviceList list = devices.get();
+      std::vector<CaptureDeviceInfo> out;
+      out.reserve(static_cast<std::size_t>(list.devices_size()));
+      for (const proto::CaptureDeviceInfo& info : list.devices()) {
+        out.push_back(fromProto(info));
+      }
+      return out;
```
alan-george-lk: Worth collapsing into a `from` helper (see below) for similar pattern/unit testability? `listDevices` has tests but am assuming most if not all will fail in CI, moving this bit out at least covers some of it

src/capture_source.cpp
```diff
@@ -0,0 +1,459 @@
...
+    if (callback) {
+      callback(result);
+    }
+  });
+
+  return source;
+}
+
+TrackPublishOptions CaptureSource::publishOptions(TrackPublishOptions options) const {
+  const TrackPublishOptions& dictated = source_publish_options_;
```
alan-george-lk: Nit: not sure the value of making a const reference to a member variable, shortens the code below a bit, but is a layer of readability indirection since it's directly accessing the variable anyways?

include/livekit/capture_source.h
```diff
@@ -0,0 +1,453 @@
...
+  TrackPublishOptions publishOptions(TrackPublishOptions options = {}) const;
+
+  /// Set the terminal notification. Set this before @ref start().
+  void setOnFinishedCallback(FinishedCallback callback);
+
+  /// Start pumping frames into the RTC video source.
+  ///
+  /// @throws CaptureSourceError if the capture was already started or the
+  ///         FFI call fails.
+  void start();
```
alan-george-lk: Consider `bool` on this and `stop()`, similar to `room.connect()/disconnect()`

src/capture_source.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+/// Rejects a resolution that cannot be represented on the wire.
+///
+/// @ref CaptureResolution is signed while the protocol carries unsigned
+/// dimensions, so a negative value would otherwise be reinterpreted as a
+/// dimension near 2^32.
+void validateResolution(const CaptureResolution& resolution, const char* field) {
+  if (resolution.width <= 0 || resolution.height <= 0) {
+    throw CaptureSourceError(std::string(field) + " must be positive, got " + std::to_string(resolution.width) + "x" +
+                             std::to_string(resolution.height));
```
alan-george-lk, line 51: Are there any upper invalid bounds to consider?

src/tests/integration/test_capture_source.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+constexpr int kMinFramesReceived = 5;
+constexpr int kCaptureWidth = 1280;
+constexpr int kCaptureHeight = 720;
+constexpr std::uint32_t kCaptureFramerateFps = 30;
+
+/// Skips the calling test when the FFI library was built without the capture
+/// feature. A feature-less FFI reports only a generic invalid handle, so this
+/// has to be decided at compile time rather than sniffed from an error string.
+#ifdef LIVEKIT_TEST_CAPTURE_ENABLED
```
alan-george-lk: Do we need a specific test flag? What about just using a `LIVEKIT_CAPTURE_ENABLED` from the public API/build, or is the idea to hide from users

include/livekit/capture_source.h
stephen-derosa: @ladvoc has this gotten into the rust layer?

src/tests/integration/test_capture_source.cpp
stephen-derosa: @ladvoc im trying to recall the conversation but i cant remember

include/livekit/capture_source.h
stephen-derosa: done

.github/workflows/tests.yml
stephen-derosa: Good callout, I added a clarifying comment. Effectively we are doing this to ensure we go through through each platform’s native package mechanism (linux/mac use latest whereas windows uses an explicit pinned version).

include/livekit/capture_source.h
stephen-derosa: nice good catch, ill align with our SDK

stephen-derosa: updated

src/capture_source.cpp
stephen-derosa: hm, i wasnt able to find upper bounds on resolution anywhere in the rust SDK (@ladvoc any idea if this exists ?) and i dont feel like its appropriate to set the precedent here. 

The resolution is currently just bounded by `INT_MAX`

include/livekit/capture_source.h
stephen-derosa: for sure, updated the comment

include/livekit/capture_source.h
stephen-derosa: no specific reason -- added

include/livekit/capture_source.h
stephen-derosa: done

include/livekit/capture_source.h
stephen-derosa: done

src/capture_source.cpp
stephen-derosa: good catch, yeah the optional is redundant because the catch returns before the future is consumed. I’ll use a plain std::future in both factory paths and move it into the mapping lambda after the request succeeds.

src/capture_source.cpp
stephen-derosa: updated

src/capture_source.cpp
stephen-derosa: yup good point, ill add the helper add some tests

stephen-derosa: also broke this into the private api

src/capture_source.cpp
stephen-derosa: fair enough, since the lambda takes in a const ref anyway. removed

include/livekit/capture_source.h
stephen-derosa: yup agreed, done

src/tests/integration/test_capture_source.cpp
stephen-derosa: good call no this should be `LIVEKIT_CAPTURE_ENABLED` specifically to cover the case where users can build with it off

include/livekit/capture_source.h
alan-george-lk: After reviewing the Rust side earlier today, looks like we have GStreamer-specific bit rates in Rust too, so maybe better to match that? CC @ladvoc on thoughts, I'm curious if both Rust and C++ can consolidate bitrates into a more generic unit/enum. Not a biggie if not, just trying to minimize API surface

src/tests/integration/test_capture_source.cpp
stephen-derosa: check if this is actually needed or should be inferred

include/livekit/capture_source.h
stephen-derosa: should be in rust, need to float up here

PR #228, opened 2026-07-30
Only build Rust livekit-ffi crate

alan-george-lk:
@ladvoc noted that this SDK was building all of Rust when only the `livekit-ffi` crate is actually used. This resolves that.


ladvoc:
Note that even if this doesn't immediately improve CI times, it will in the case of cache invalidation and for clean local builds.

stephen-derosa (approved):

PR #229, opened 2026-07-31
Build explicit Ubuntu versions, add Ubuntu 22.04 support

alan-george-lk:
## Summary

Adjust Linux release to build against explicit Ubuntu 22.04 and 24.04 runners instead of `ubuntu-latest`, and publish explicit release artifacts:

New: `livekit-sdk-ubuntu-22.04-{x64,arm64}` and `livekit-sdk-ubuntu-24.04-{x64,arm64}`
Deprecated (kept for backwards compatibility): `livekit-sdk-linux-{x64,arm64}` (simply repackaged from the existing Ubuntu 24.04 builds)

Also fixes Rust `target/` cache keys to use matrix.os rather than runner.os, so 24.04-compiled artifacts are not restored on 22.04 runners (which caused GLIBC_2.39 failures). Tests and builds now run on both Ubuntu versions.

## Testing

- Verified in ROS Portal branch against release candidate
- Verified in C++ Example collection against old deprecated "linux" repackaged artifact

## Context

In [ROS Portal](https://github.com/livekit/ros-portal), ROS Humble targets Ubuntu 22.04, and the "linux" LiveKit SDK targets 24.04. The glibc version for 24.04 was not backwards compat on 22.04, but the 22.04 glibc is forwards compatible in 24.04. Thus, we were incurring 30+min SDK build times in ROS Portal CI to rebuild the LiveKit SDK against 22.04. 

Therefore in ROS Portal, all LiveKit SDK links will be against 22.04 based artifacts, but will work in 22.04, 24.04, and 26.04 Ubuntu runtime versions for the respective distros.

This approach is also what FoxGlove takes for their SDK/ROS bridge equivalent. [Link](https://github.com/foxglove/foxglove-sdk/blob/main/.github/workflows/draft-ros-release.yml#L23-L24)

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -33,11 +33,19 @@ jobs:
       matrix:
         include:
           - os: ubuntu-22.04
-            name: linux-x64
+            name: ubuntu-22.04-x64
             build_cmd: ./build.sh release-examples
             build_dir: build-release
           - os: ubuntu-22.04-arm
-            name: linux-arm64
+            name: ubuntu-22.04-arm64
+            build_cmd: ./build.sh release-examples
+            build_dir: build-release
+          - os: ubuntu-24.04
+            name: ubuntu-24.04-x64
+            build_cmd: ./build.sh release-examples
+            build_dir: build-release
+          - os: ubuntu-24.04-arm
+            name: ubuntu-24.04-arm64
             build_cmd: ./build.sh release-examples
             build_dir: build-release
           - os: macos-26-xlarge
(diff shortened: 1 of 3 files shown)


.github/workflows/builds.yml
```diff
@@ -161,14 +169,19 @@ jobs:
       # step that runs AFTER the "Clean after build" cleanup wipes target/debug
       # and target/release, which is why the old cache was always ~empty. The
       # restore/save split lets us snapshot the populated target/ before cleanup.
+      #
+      # Key on matrix.os (runner image), not runner.os. runner.os is just
```
stephen-derosa: is this comment important or is it more just a note by AI about the diff that could be removed?

.github/workflows/make-release.yml
```diff
@@ -413,15 +422,30 @@ jobs:
...
+          for platform in ubuntu-22.04-x64 ubuntu-22.04-arm64 ubuntu-24.04-x64 ubuntu-24.04-arm64 macos-arm64 macos-x64; do
             dirName="livekit-sdk-${platform}-${VERSION}"
             if [[ -d "${dirName}" ]]; then
               echo "Creating archive for ${platform}..."
               tar -czf "${{ github.workspace }}/release-assets/${dirName}.tar.gz" "${dirName}"
               echo "Created: ${dirName}.tar.gz"
             fi
           done
+
+          # Deprecated compatibility aliases for existing Linux release consumers.
```
stephen-derosa, line 434: nice!

stephen-derosa (approved):
small questions, LGTM.

Surprised we didnt need any Cmake/deps changes in the repo -- nice!

.github/workflows/builds.yml
alan-george-lk: From AI, I also flagged it in the self review, it is worth noting mentally since we have multiple linux builds now so matrix is needed, but is it worth a comment? Perhaps not. I'll remove

PR #234, opened 2026-08-05
other_sdks integration

stephen-derosa:
# Overview
expose other_sdks for apps like `ros-portal`

### DIFF
--- a/include/livekit/room.h
+++ b/include/livekit/room.h
@@ -121,6 +121,13 @@ struct RoomOptions {
   ///
   /// If unset, the Rust SDK default is used.
   std::optional<std::chrono::milliseconds> connect_timeout;
+
+  /// Additional LiveKit SDKs layered on top of this one, reported to the server
+  /// as part of the client info.
+  ///
+  /// Comma separated list of `name:version` pairs, e.g.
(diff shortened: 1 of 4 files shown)


include/livekit/room.h
```diff
@@ -121,6 +121,13 @@ struct RoomOptions {
   ///
   /// If unset, the Rust SDK default is used.
   std::optional<std::chrono::milliseconds> connect_timeout;
+
+  /// Additional LiveKit SDKs layered on top of this one, reported to the server
+  /// as part of the client info.
+  ///
+  /// Comma separated list of `name:version` pairs, e.g.
+  /// "ros_portal:1.2.3,components-cpp:2.0.0". If unset, no additional SDKs are reported.
+  std::optional<std::string> other_sdks;
```
ladvoc, line 132: suggestion(non-blocking): I would add a comment that this is for internal use and/or hide from generated docs if possible like we do in Rust.

ladvoc (approved):

include/livekit/room.h
stephen-derosa: done

xianshijing-lk (approved):

PR #240, opened 2026-08-20
Consolidate token helper script

alan-george-lk:
## Summary
- Consolidate token helpers into `scripts/set-test-tokens.sh`
  - Standardize the script name on hyphen-separated shell-script naming
- Remove obsolete `.token_helpers` folder, scripts, and README
- Update CI, test guidance, documentation, and integration-test comments to use the new path


stephen-derosa (approved):
lgtm -- IIRC the original idea was that these would only be used internally, but theres no real reason other devs cant use them

PR #244, opened 2026-08-25
Remove Docker image building

alan-george-lk:
Removes Docker image building and GHCR pushing. Was decided internally that the build times (20-40min) on release weren't worth it, and stats showed negligible (internal downloads) usage if any.

If you are a user that depended on this, consider integrating into your own Dockerfile or raise a GitHub issue otherwise.


stephen-derosa (approved):
hello fast build times

PR #249, opened 2026-09-02
UniFFI C++ Bindgen Initial Integration

alan-george-lk:
## Summary

This PR introduces the initial steps for `uniffi` integration. Depends on https://github.com/livekit/rust-sdks/pull/1389 and release tag.

For this initial integration, a private header/adapter translation unit is added to test the UniFFI build version signature offered by the current `livekit-ffi`. This approach was taken to:

1. Not expose this signature in the public header
2. Not change building or linking behavior to expose the bindings directly to the unit test (better tests the build/link this way anyway)
3. Allows these internal files to be deleted later once actual critical-path SDK features are moved over to UniFFI

## Testing

Significant testing occurred locally across the bingden fork repository (testing C++ features), and in this repository through building/linking/etc. The new unit test is isolated and can be deleted later once other features move to UniFFI as mentioned above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,6 +269,10 @@ endif()
 message(STATUS \"[run_cargo.cmake] CFG=\${CFG}  CARGO=\${CARGO}  PROTOC=\${PROTOC_PATH}\")
 set(ENV{PROTOC} \"\${PROTOC_PATH}\")
 
+if(DEFINED CARGO_TARGET_DIR AND NOT CARGO_TARGET_DIR STREQUAL \"\")
+  set(ENV{CARGO_TARGET_DIR} \"\${CARGO_TARGET_DIR}\")
+endif()
+
 if(DEFINED GCC_LIB_DIR AND NOT GCC_LIB_DIR STREQUAL \"\")
   set(ENV{RUSTFLAGS} \"-L \${GCC_LIB_DIR} \$ENV{RUSTFLAGS}\")
   set(ENV{LD_LIBRARY_PATH} \"\${GCC_LIB_DIR}:\$ENV{LD_LIBRARY_PATH}\")
@@ -368,6 +372,8 @@ add_custom_target(build_rust_ffi
   DEPENDS "${RUST_LIB_DEBUG}" "${RUST_LIB_RELEASE}"
 )
 
+include(uniffi_cpp)
+
 # Note: protozero_plugin.o removal is no longer needed since we use dynamic libraries on Unix
 
 add_library(livekit SHARED
@@ -466,6 +472,7 @@ target_link_libraries(livekit
   PRIVATE
     spdlog::spdlog
     nlohmann_json::nlohmann_json
+    livekit_uniffi_cpp
     livekit_ffi
     ${LIVEKIT_PROTOBUF_TARGET}
 )
--- a/cmake/uniffi_cpp.cmake
+++ b/cmake/uniffi_cpp.cmake
@@ -67,6 +67,9 @@ add_custom_command(
 )
 add_custom_target(build_livekit_ffi_metadata
   DEPENDS "${LIVEKIT_UNIFFI_METADATA_LIBRARY}")
+# Both commands invoke rustup/Cargo and share toolchain state. Keep them
+# serialized so fresh CI runners cannot race while installing the toolchain.
+add_dependencies(build_livekit_ffi_metadata build_rust_ffi)
 
 add_custom_command(
   OUTPUT
--- a/cmake/uniffi_cpp.cmake
+++ b/cmake/uniffi_cpp.cmake
@@ -20,17 +20,13 @@ set(LIVEKIT_UNIFFI_CPP_HEADER
 set(LIVEKIT_UNIFFI_CPP_SCAFFOLDING_HEADER
     "${LIVEKIT_UNIFFI_CPP_GENERATED_DIR}/livekit_ffi_scaffolding.hpp")
 
-# Generate from an unstripped host library. Release Linux builds strip the
-# UniFFI metadata symbols, while generated C++ source is platform-independent.
-set(LIVEKIT_UNIFFI_METADATA_TARGET_DIR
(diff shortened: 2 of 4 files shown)


src/tests/CMakeLists.txt
```diff
@@ -89,6 +89,8 @@ if(UNIT_TEST_SOURCES)
   target_link_libraries(livekit_unit_tests
     PRIVATE
       livekit
+      livekit_uniffi_cpp
```
alan-george-lk, line 92: These are only needed until we actually get features through the SDK in UniFFI. I can add a comment for this

cmake/uniffi_cpp.cmake
stephen-derosa, line 1: dig that this is its own cmake

CMakeLists.txt
```diff
@@ -445,7 +453,14 @@ if(UNIX AND NOT APPLE)
 endif()
 
 
-target_sources(livekit PRIVATE $<TARGET_OBJECTS:livekit_proto>)
+target_sources(livekit PRIVATE
+  $<TARGET_OBJECTS:livekit_proto>
+  $<TARGET_OBJECTS:livekit_uniffi_cpp>
+)
+set_source_files_properties(
```
stephen-derosa, line 460: TIL!

CMakeLists.txt
```diff
@@ -401,6 +408,7 @@ add_library(livekit SHARED
   src/token_source_json.cpp
   src/token_source_jwt.cpp
   src/token_source_internal.h
+  src/uniffi_bindgen_adapter.cpp
```
stephen-derosa, line 411: should we be putting uniffi bindgen into its own lib which livekit links against? Maybe overkill but im thinking it will allow isolated development of the entire uniffi interfaces.

src/uniffi_bindgen_adapter.cpp
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#include "uniffi_bindgen_adapter.h"
+
+#include "livekit_ffi.hpp"
+
+namespace livekit {
+
+std::optional<std::string> uniffiBindgenBuildVersion() {
+  auto version = livekit_ffi::build_version();
```
stephen-derosa, line 24: is `livekit_ffi::build_version()` a rust impl thats getting called in cpp?

stephen-derosa: i.e. is this the "Secret Sauce"

alan-george-lk: This is currently the only-available `uniffi` exported function in `livekit-ffi`: https://github.com/livekit/rust-sdks/blob/main/livekit-ffi/src/build_info.rs#L20-L23

This is the secret sauce in the sense that `livekit_ffi` is the uniffi generated namespace, and `build_version` is rust impl interfaced by the bindgen, so yes.

I tested all of this infra against `livekit-uniffi` but that is not actually targeted for critical path (per @1egoman), but allowed me to test more things:

https://github.com/livekit/rust-sdks/blob/main/livekit-uniffi/src/access_token.rs#L185-L189

and

https://github.com/livekit/rust-sdks/blob/main/livekit-uniffi/src/log_forward.rs#L27-L32

(removed from this branch for PR but part of history)

CMakeLists.txt
alan-george-lk: Replied internally, but for viz the idea is to keep `livekit_ffi.so/.dylib` and migrate the implementation, so the `uniffi` stuff is just an internal linkage via CMake `OBJECT`

src/uniffi_bindgen_adapter.cpp
stephen-derosa: sweet thanks!

stephen-derosa (approved):

src/uniffi_bindgen_adapter.cpp
xianshijing-lk: shouldn't we expoe the ffi version via the public livekit header instead ? 

livekit_ffi.hpp seems internal to me.

src/uniffi_bindgen_adapter.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "livekit/visibility.h"
+
+namespace livekit {
+
+/// @brief Gets the build version through the generated UniFFI binding.
+///
+/// @return The generated binding's build version, or no value if it is invalid
+/// or the binding call fails.
+LIVEKIT_INTERNAL_API std::optional<std::string> uniffiBindgenBuildVersion();
```
xianshijing-lk, line 36: same here, the uniffi_bindgen_adapter.h seems internal to me, can we wrap it via some livekit public interface rather than directly expose it to public ?

alan-george-lk: Our build versions come from a compile-time generated file, with no actual linkage or implementation (just static values in a header): https://github.com/livekit/client-sdk-cpp/blob/main/build.h.in

I considered integrating this into the header, but that would significantly change how that file works and the build code around it. The goal of this integration was the infra around the bindgen fork, the Rust bindgen calls, and the CMake to wire it all up. This adapter is intentionally temporary and isn't actually useful, I planned on removing it later.

Hope this makes sense.

alan-george-lk: There might be a way to compile the bindgen code into a simple binary that CMake could call to get the same file statically, but that didn't feel useful

xianshijing-lk: step back a bit, what is the purpose of exposing this uniffiBindgenBuildVersion() publicly ? is it for testing purpose ?

xianshijing-lk (approved):
Sorry, I missed read the macro, it is LIVEKIT_INTERNAL_API

src/uniffi_bindgen_adapter.h
alan-george-lk: Resolved internally -- not exposed publicly and will be removed once actual features migrate to `uniffi`. if we don't go this route and test directly in unit/integration tests, we would need to change the linking behavior which I was trying to avoid.

PR #256, opened 2026-09-08
set-test-tokens: use env vars as default for secret/key/url

stephen-derosa:


scripts/set-test-tokens.sh
```diff
@@ -53,12 +53,14 @@ if [[ $# -ne 0 ]]; then
...
-LIVEKIT_API_KEY="devkey"
-LIVEKIT_API_SECRET="secret"
+LIVEKIT_API_KEY="${LIVEKIT_API_KEY:-devkey}"
+LIVEKIT_API_SECRET="${LIVEKIT_API_SECRET:-secret}"
+LIVEKIT_URL="${LIVEKIT_URL:-ws://localhost:7880}"
 LIVEKIT_VALID_FOR="99999h"
-LIVEKIT_URL="ws://localhost:7880"
 _grant_json='{"canPublish":true,"canSubscribe":true,"canPublishData":true}'
 
+echo "Using LIVEKIT_URL: $LIVEKIT_URL"
```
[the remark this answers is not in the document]
stephen-derosa: done

scripts/set-test-tokens.sh
```diff
@@ -53,12 +53,14 @@ if [[ $# -ne 0 ]]; then
...
-LIVEKIT_API_KEY="devkey"
-LIVEKIT_API_SECRET="secret"
+LIVEKIT_API_KEY="${LIVEKIT_API_KEY:-devkey}"
+LIVEKIT_API_SECRET="${LIVEKIT_API_SECRET:-secret}"
+LIVEKIT_URL="${LIVEKIT_URL:-ws://localhost:7880}"
 LIVEKIT_VALID_FOR="99999h"
-LIVEKIT_URL="ws://localhost:7880"
 _grant_json='{"canPublish":true,"canSubscribe":true,"canPublishData":true}'
 
+echo "set_test_tokens: Using LIVEKIT_URL: $LIVEKIT_URL" >&2
```
alan-george-lk, line 62: Nit: Can we avoid printing here? We have plenty of logs across SDK layers/end products (ROS Portal) that log which URL is being connected to. I also run individual tests often and call into this script first, so I'll be getting more output

alan-george-lk (approved):

scripts/set-test-tokens.sh
stephen-derosa: I think we need _some_ log to indicate the creds being used so we can prevent devs from using the wrong creds. I could change it to "using local server" or "using cloud server" ?

alan-george-lk: Fair idea, I took the approach in `ros-portal` to print at the token-helper level, wonder if we should just move these logs into the token source SDK code itself, I think that would be the best of all worlds

https://github.com/livekit/ros-portal/blob/main/src/ros_portal/src/token_loader.cpp#L69-L77

